PR #2, opened 2017-04-10
Upcoming assessments list

apalaniuk:
Super thin slice of upcoming assessments list.

* Currently uses built-in mocked data
* Styling likely doesn't 100% match any specific design as no recent desktop design, so generally follows current mobile design; general concepts (RTL/lang terms) are considered


lang/en.html
```diff
@@ -13,6 +13,8 @@
 	*/
 	window.D2L.UpcomingAssessments.LocalizeBehavior.LangEnBehavior = {
 		en: {
+			'upcomingAssessments': 'Upcoming Assessments',
+			'dueDate': 'Due: {dueDate}'
```
apalaniuk: Should probably use real date formatting here, but need to check API results and how things need to be converted, so with the current test data it doesn't really match designs.

d2l-upcoming-assessments.html
```diff
@@ -14,18 +16,87 @@
...
+					'dueDate': '2017-04-08',
+					'isCompleted': true
+				}];
+			},
+
+			_onResponse: function(response) {
+
+				if (response.detail.status === 200) {
+
+					this.set('assessments', response.detail.xhr.response);
```
cgalvind2l: What if the request fails?

lang/en.html
cgalvind2l: :+1:

test/d2l-assessments-list-item.html
```diff
@@ -0,0 +1,88 @@
...
+	</head>
+	<body>
+
+		<test-fixture id="basic">
+			<template>
+				<d2l-assessments-list-item></d2l-assessments-list-item>
+			</template>
+		</test-fixture>
+
+		<script>
```
cgalvind2l: Why not break the script into its own file like we usually do?

test/d2l-assessments-list.html
```diff
@@ -0,0 +1,79 @@
...
+
+						element.set('assessmentItems', assessmentItems);
+						element.$$('template').render();
+
+						var listElements = Polymer.dom(element.root).querySelectorAll('d2l-assessments-list-item');
+
+						expect(listElements.length).to.equal(assessmentItems.length);
+					});
+
+				});
```
cgalvind2l: May be worth adding a test for the behaviour when an item is changed.

d2l-assessments-list-item.html
```diff
@@ -0,0 +1,107 @@
...
+				line-height: 1;
+			}
+
+		</style>
+			<div class="info-container">
+				<h3 class="assessment-title">{{assessmentItem.title}}</h3>
+				<div class="assessment-info">
+					<div class="course-name">{{assessmentItem.courseName}}</div>
+					<iron-icon class="separator-icon" icon="d2l-tier1:bullet"></iron-icon>
+					<div class="item-type">{{assessmentItem.itemType}}</div>
```
apalaniuk: That's to be left to when it's hooked up to the real API

d2l-assessments-list-item.html
```diff
@@ -0,0 +1,107 @@
...
+				</div>
+				<div class="assessment-info due-date">{{localize('dueDate', 'dueDate', assessmentItem.dueDate)}}</div>
```
apalaniuk: There are actually a few designs - some have it with each item, some as a header that groups the following items. That will be in the follow-up PR where things happen and Bob reviews it

d2l-assessments-list.html
```diff
@@ -0,0 +1,78 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="d2l-assessments-list-item.html">
+<link href="https://fonts.googleapis.com/css?family=Lato" rel="stylesheet">
```
apalaniuk: Me too. That was due to some demo page weirdness overwriting fonts. Should definitely be taken out.

test/d2l-assessments-list-item.html
apalaniuk: This was created through `polymer init` scaffolding - I can break this out now if it makes sense (was thinking if it gets too large instead, but that might be a not-so-fun PR)

d2l-assessments-list-item.html
```diff
@@ -0,0 +1,107 @@
+<link rel="import" href="../polymer/polymer.html">
+<link rel="import" href="../iron-icon/iron-icon.html">
+<link rel="import" href="../d2l-typography/d2l-typography-shared-styles.html">
+<link rel="import" href="../d2l-colors/d2l-colors.html">
+<link rel="import" href="../d2l-icons/tier1-icons.html">
+<link rel="import" href="localize-behavior.html">
+
+
+<!--
```
apalaniuk: Autogenerated scaffolding stuff that I can totally take out

d2l-upcoming-assessments.html
apalaniuk: It's a good point. I don't think there are any designs or guidance around this so I'll ask Bob. The endpoint currently isn't actually used - but, for now I should throw in an error message of some type.

d2l-assessments-list-item.html
cgalvind2l: Not a big deal, just wondering since we don't have it anywhere else in the work we do.

test/d2l-assessments-list-item.html
cgalvind2l: Yeah I think it would be worthwhile to follow our usual pattern even if with small files it makes less sense.

d2l-upcoming-assessments.html
```diff
@@ -1,31 +1,104 @@
...
+				}];
+			},
+
+			_onResponse: function(response) {
+
+				if (response.detail.status === 200) {
+					this.set('assessments', response.detail.xhr.response);
+					this.notifyPath('assessments.*');
+				} else {
+					this.set('_showRequestFailedError', true);
```
cgalvind2l: Are you sure that `iron-ajax` works like this? Josh and I saw a problem earlier today where, on an error, `iron-ajax` doesn't call its `on-response` but wants to have an `on-error` instead.

apalaniuk: I wanted to blame some different `d2l-ajax` convention on this, but nope, they all use separate handlers, so.. yeah. Pretty sure the "real" error handling tests I was about to write to catch this.

test/d2l-assessments-list.js
```diff
@@ -0,0 +1,84 @@
...
+			element.$$('template').render();
+
+			element.set('assessmentItems', newAssessmentItems);
+			element.$$('template').render();
+
+			var listElements = Polymer.dom(element.root).querySelectorAll('d2l-assessments-list-item');
+
+			expect(listElements.length).to.equal(newAssessmentItems.length);
+		});
+
```
cgalvind2l: Would it be worth having a test that ensures that the grouping structure is what you expect it to be?

cgalvind2l:
Might be good to have a quick test of the localization behaviour as well

test/d2l-assessments-list.js
apalaniuk: I was thinking this too, especially because I'm still not sure about what guarantees exist in terms of ordering from the existing dates API. I might pull any semblance of grouping logic out of this PR so when it's added in the next one that hooks things up and will be design reviewed (since things are *sorta* in flux at this point), it's tested when it's pulled in.

d2l-upcoming-assessments.html
apalaniuk: On second thought, this is valid for receiving a non-200 response and should stay; There does need to be an additional handler for other errors like timeouts, etc.

d2l-upcoming-assessments.html
```diff
@@ -1,31 +1,111 @@
...
+			}
 		</style>
-		<h2>Upcoming Assessments</h2>
+
+		<d2l-ajax
+			auto
+			url="{{endpoint}}"
+			handle-as="json"
+			on-iron-ajax-response="_onResponse"
+			on-iron-ajax-error="_onError">
```
cgalvind2l: Alternatively, `on-error` could also go to the `_onResponse` function, since it will handle it properly.

apalaniuk: Looks like this is true for all cases - I wasn't sure if there was any case where an error could occur without `response.detail` being available, but it looks like it always should be.

cgalvind2l:
LGTM

PR #4, opened 2017-06-21
use frau-ci to auto-increment the version and tag

CodeBaboon:


CodeBaboon:
Assuming this doesn't break the pr/commit builds the only real way to test this is to merge it and see what happens.

PR #8, opened 2017-06-21
straws and the grasping of them

CodeBaboon:


CodeBaboon:
Upside: I've at least just learned that I can see the issue in the View Config tab on the Travis build, even ones from within the PR, so I don't have to keep merging to figure out if I've fixed the issue.

Downside: Travis hates me.

PR #10, opened 2017-06-22
Empty states

apalaniuk:
Add empty/error states, break some things out, etc. Caveats:

* Uses `iron-ajax` in, whereas we'll almost certainly want to use `d2l-fetch` instead, but decided to try to keep this simple(r) for now. As a result, there are potential race conditions since there are two endpoints with independent requests setting a global error state, but that will be solved when a _real_ workflow exists.

* Since the "assessments URL" (should probably be _activities URL_? Will rename when the real API is hooked up, if necessary, since it isn't set in the consumer yet anyway) isn't implemented, you're just going to get the "no upcoming assignments" message rather than a real error (except on the demo page). Again, that's corrected when a real workflow is implemented.

* As per designer, we're not concerned about not having a given name available from the users API, yet. This pattern exists in parent-portal-ui as well, so can be solved at the same time if it's an issue.

* Uses some Polymer 1.x helpers that won't exist in the future, but they're easily discoverable with global search, so.

* Probably some other things I'm not thinking of. (Like, the pre-data state will in the future be handled by parent-portal-ui skeleton, rather than being empty)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,6 @@ dist: trusty
 sudo: required
 before_script:
 - npm run test:lint:js
-- npm run test:lint:wc
 script:
 - xvfb-run wct
 - if [ "$TRAVIS_BRANCH" == "master" ]; then frauci-update-version; fi
--- a/bower.json
+++ b/bower.json
@@ -5,16 +5,15 @@
   "dependencies": {
     "app-localize-behavior": "PolymerElements/app-localize-behavior#^1.0.2",
     "d2l-colors": "^2.2.3",
-    "d2l-hm-constants-behavior": "git://github.com/Brightspace/d2l-hm-constants-behavior.git#^1.0.8",
+    "d2l-hm-constants-behavior": "Brightspace/d2l-hm-constants-behavior#^1.0.8",
     "d2l-icons": "^3.1.2",
     "d2l-typography": "^5.3.0",
     "fetch": "^2.0.3",
     "iron-ajax": "PolymerElements/iron-ajax#^1.4.4",
     "iron-icon": "^1.0.12",
-    "polymer": "Polymer/polymer#^1.6.0"
+    "polymer": "Polymer/polymer#^1.9.2"
   },
   "devDependencies": {
-    "web-component-tester": "^4.0.0",
     "webcomponentsjs": "webcomponents/webcomponentsjs#^0.7.0",
     "d2l-demo-template": "~0.0.11"
   }
--- a/bower.json
+++ b/bower.json
@@ -3,15 +3,15 @@
   "description": "Widget to display upcoming assessments",
   "main": "d2l-upcoming-assessments.html",
   "dependencies": {
-    "polymer": "Polymer/polymer#^1.6.0",
-    "app-localize-behavior": "PolymerElements/app-localize-behavior#~0.10.0",
-    "d2l-ajax": "^3.3.0",
+    "app-localize-behavior": "PolymerElements/app-localize-behavior#^1.0.2",
     "d2l-colors": "^2.2.3",
     "d2l-hm-constants-behavior": "git://github.com/Brightspace/d2l-hm-constants-behavior.git#^1.0.8",
-    "d2l-typography": "^5.3.0",
     "d2l-icons": "^3.1.2",
+    "d2l-typography": "^5.3.0",
     "fetch": "^2.0.3",
-    "iron-icon": "^1.0.12"
+    "iron-ajax": "^2.0.2",
+    "iron-icon": "^1.0.12",
+    "polymer": "Polymer/polymer#^1.6.0"
   },
   "devDependencies": {
     "web-component-tester": "^4.0.0",
--- a/bower.json
+++ b/bower.json
@@ -8,7 +8,6 @@
     "d2l-hm-constants-behavior": "Brightspace/d2l-hm-constants-behavior#^1.0.8",
     "d2l-icons": "^3.1.2",
     "d2l-typography": "^5.3.0",
-    "fetch": "^2.0.3",
     "iron-ajax": "PolymerElements/iron-ajax#^1.4.4",
     "iron-icon": "^1.0.12",
     "polymer": "Polymer/polymer#^1.9.2"
--- a/bower.json
+++ b/bower.json
@@ -9,7 +9,7 @@
     "d2l-icons": "^3.1.2",
     "d2l-typography": "^5.3.0",
     "fetch": "^2.0.3",
-    "iron-ajax": "^2.0.2",
+    "iron-ajax": "PolymerElements/iron-ajax#^1.4.4",
     "iron-icon": "^1.0.12",
     "polymer": "Polymer/polymer#^1.6.0"
   },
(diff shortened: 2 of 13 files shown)


.travis.yml
```diff
@@ -4,7 +4,6 @@ dist: trusty
 sudo: required
 before_script:
 - npm run test:lint:js
-- npm run test:lint:wc
```
apalaniuk: Absolute URLs break polylint it seems.

ryantmer: Ah, I think this was fixed with `polymer lint`. Polylint the separate package is a bit out of date I believe. (`polymer-cli` uses `polymer-linter` instead)

bower.json
```diff
@@ -3,13 +3,15 @@
...
     "d2l-colors": "^2.2.3",
-    "d2l-typography": "^5.3.0",
+    "d2l-hm-constants-behavior": "git://github.com/Brightspace/d2l-hm-constants-behavior.git#^1.0.8",
     "d2l-icons": "^3.1.2",
-    "iron-icon": "^1.0.12"
+    "d2l-typography": "^5.3.0",
+    "fetch": "^2.0.3",
+    "iron-ajax": "PolymerElements/iron-ajax#^1.4.4",
+    "iron-icon": "^1.0.12",
+    "polymer": "Polymer/polymer#^1.6.0"
```
ryantmer: Mind bumping to 1.9.2? It's the most recent non-2.0 release.

bower.json
```diff
@@ -3,13 +3,15 @@
   "description": "Widget to display upcoming assessments",
   "main": "d2l-upcoming-assessments.html",
   "dependencies": {
-    "polymer": "Polymer/polymer#^1.6.0",
-    "app-localize-behavior": "PolymerElements/app-localize-behavior#~0.10.0",
-    "d2l-ajax": "^3.3.0",
+    "app-localize-behavior": "PolymerElements/app-localize-behavior#^1.0.2",
     "d2l-colors": "^2.2.3",
-    "d2l-typography": "^5.3.0",
+    "d2l-hm-constants-behavior": "git://github.com/Brightspace/d2l-hm-constants-behavior.git#^1.0.8",
```
ryantmer: This has bitten me once or twice, but if you leave out the git://github.com, your client will install via whichever protocol it prefers :)

components/d2l-upcoming-assessments.html
```diff
@@ -20,88 +21,126 @@
...
 
-		<d2l-ajax
-			auto
-			url="{{endpoint}}"
+		<iron-ajax
+			id="assessmentsRequest"
+			url="[[assessmentsUrl]]"
+			handle-as="json"
+			on-response="_onAssessmentsResponse"
+			on-error="_onError">
```
ryantmer: iirc `on-error` (or maybe `on-response`?) can cause some issues (in... FF maybe?), hence the introduction of the more-specific-and-less-likely-to-collide `on-iron-ajax-*` events.

components/d2l-upcoming-assessments.html
```diff
@@ -20,88 +21,126 @@
 				margin-left: 0px;
 				margin-right: 15px;
 			}
+
+			.error-message, .no-upcoming-assessments {
+				@apply --d2l-body-compact-text;
```
ryantmer: Huh, if I'm not mistaken the `d2l-typography-shared-styles` import wasn't required before this. Nice for it to already be there for you 😆

components/d2l-upcoming-assessments.html
```diff
@@ -20,88 +21,126 @@
...
+			observers: [
+				'_onAssessmentsUrlChanged(assessmentsUrl, token)',
+				'_onUserUrlChanged(userUrl, token)'
+			],
+
+			_onAssessmentsResponse: function(response) {
+				if (response.detail.status === 200 || response.detail.status === 304) {
+					this._showError = false;
+					this.set('_assessments', response.detail.xhr.response);
+					this.notifyPath('_assessments.*');
```
ryantmer: Curious why this is needed after the `this.set`?

components/d2l-upcoming-assessments.html
```diff
@@ -20,88 +21,126 @@
...
+			_onAssessmentsUrlChanged: function() {
+				this.debounce('getAssessmentsInfo', this._getAssessmentsInfo, 250);
+			},
+
+			_onUserUrlChanged: function() {
+				this.debounce('getUserInfo', this._getUserInfo, 250);
+			},
+
+			_hasUpcomingAssessments: function() {
+				return this._assessments && this._assessments.length;
```
ryantmer: Should have a `> 0` on the length check I do believe (since an empty array would still return truthy for the above, although... idk if that's even possible in this case)

test/d2l-upcoming-assessments.js
```diff
@@ -40,14 +43,15 @@ describe('<d2l-upcoming-assessments>', function() {
 
 			element = fixture('valid-endpoint');
 
-			setTimeout(function() {
+			clock.tick(500);
+
+			setTimeout(function(done) {
```
ryantmer: Don't need to pass `done` here afaik (same with test below)

bower.json
apalaniuk: Yeah, I remember pushing for that during d2l-formsgate. I grabbed this directly from parent-portal-ui, so I'll change it in both areas I think.

components/d2l-upcoming-assessments.html
apalaniuk: Good catch. I interpreted those were wrapper events implemented on `d2l-ajax` and quickly checked `iron-ajax`'s docs, which use the `on-request` style. Thanks!

components/d2l-upcoming-assessments.html
apalaniuk: 0 is certainly falsy, though I don't have a problem in being more explicit and removing all doubt.

components/d2l-upcoming-assessments.html
apalaniuk: I... don't remember. Whatever the reason, it seems vestigial at best now.

components/d2l-upcoming-assessments.html
apalaniuk: I was thinking ahead 🙄

test/d2l-upcoming-assessments.js
apalaniuk: If you check the last commit, you can see the exact point where I went to lunch after getting it to pass.. and never looking back. 😆

components/d2l-upcoming-assessments.html
ryantmer: o dang, yeah that's completely true as well. Although as evidenced by my (frankly embarrassing) statement above, explicit is nice 😆

CodeBaboon:
Code LGTM, will test now

CodeBaboon:
Seems like the alignment of the empty states widget is perhaps incorrect. Here is an image of the mock (notice my super straight red line):
mock-alignment

And here is what I see in the app (again with super helpful red line):
alignment

CodeBaboon:
Testing notes:

Alignment issue noted above aside, both the empty states and error states views occur correctly and appear to match the mocks (now that the no upcoming assignments message has been made less fun).

Tested on:
- [x] Chrome
- [x] FF
- [x] Edge

Could not test on mobile due to remote connection issues.

A11y:
- [x] Currently there are no interactive elements and so it is not possible to tab into the widget. NVDA does not read out the existence of the widget at all unless I use the NVDA+↓ command to read the entire page. Upcoming Assignments does appear in my Headings list when using NVDA+F7. This is probably all acceptable for now.

PR #11, opened 2017-06-27
Add all assignments overlay

CodeBaboon:
[US87820 - View all upcoming assignments empty list page](https://rally1.rallydev.com/#/128188728532d/detail/userstory/128218236208)

Also see https://github.com/Brightspace/parent-portal-ui/pull/219

**Acceptance Criteria:**
Add link to "View All Upcoming" in widget
Launch "View all Upcoming" page

NOTE: The current implementation does not match the header design for the overlay, seen here: https://app.zeplin.io/project/58d138705fdb5c5361c515df/screen/594c04f0dd10648972532c66

The reason is that modifications need to be made to the [Simple Overlay](https://github.com/Brightspace/simple-overlay) to support slotted content in the header, but we don't want to hold up other team members from performing tasks on this screen (such as using data and stuff) while that is done. So to support parallelization a follow up PR will be done to fix the header to match the mock.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -7,6 +7,8 @@
     "d2l-colors": "^2.2.3",
     "d2l-hm-constants-behavior": "Brightspace/d2l-hm-constants-behavior#^1.0.8",
     "d2l-icons": "^3.1.2",
+    "d2l-link": "^3.5.1",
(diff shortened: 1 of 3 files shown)


apalaniuk:
Code looks good, will test.

PR #12, opened 2017-06-28
US87818 - Hook up API/Display first four upcoming assignments

apalaniuk:
Requires:
https://github.com/Brightspace/parent-portal-ui/pull/221
https://github.com/Brightspace/d2l-hm-constants-behavior#master

Given a user URL, will retrieve assignments with due dates and display up to four upcoming assignment (assignments due within the next 14 calendar days - https://www.evernote.com/shard/s203/sh/e49ece72-d133-42da-b783-891d0553378d/a8962e3f10a69408).

Todo:

- [x] Don't show hidden assignments (until API updated to not include them)
- [x] Make display of messages/assignments more atomic
- [x] Investigate ~iOS~ Safari display issues


components/d2l-assessments-list-item.html
```diff
@@ -88,6 +85,24 @@ <h3 class="assessment-title">{{assessmentItem.title}}</h3>
 
 		_isCompleted: function(item) {
 			return item.isCompleted;
+		},
+
+		_getDateString: function(dueDateUTC) {
+			var msPerDay = 86400000;
+			var activityDueDateTime = new Date(dueDateUTC);
+			var startOfToday = new Date().setHours(0, 0, 0, 0);
```
CodeBaboon: I think we can have a mismatch here because the `startOfToday` is going to be based on local time but the `activityDueDateTime` is in UTC.

components/d2l-upcoming-assessments.html
```diff
@@ -87,62 +75,159 @@
...
+					})
+					.then(function(activities) {
+						return self._getActivityUsagesInfo(activities);
+					})
+					.then(function(activities) {
+						self._updateActivitiesInfo(activities);
+					})
+					.catch(function() {
+						self._onError();
+					});
```
CodeBaboon: I think you can do:

```
return this._fetchEntity(self.userUrl)
    .then(self._getActivityUsages)
    .then(self._getActivityUsagesInfo)
    .then(self._updateActivitiesInfo)
    .catch(self._onError);
```

components/d2l-upcoming-assessments.html
```diff
@@ -87,62 +75,159 @@
...
 
-			_onError: function() {
-				this._showError = true;
+			_getActivityUsagesInfo: function(activities) {
+				var requests = [];
+				var orgPromiseMap = new Map();
+				var self = this;
+
+				activities.getSubEntitiesByClass('user-assignment-activity').forEach(function(assignment) {
+					if (assignment.getSubEntityByClass('due-date')) {
```
CodeBaboon: It might be worth inverting this to be:

`if (!assignment.getSubEntityByClass('due-date')) return;`

components/d2l-upcoming-assessments.html
```diff
@@ -87,62 +75,159 @@
...
+				activities = activities
+					.filter(function(activity) {
+						return activity.dueDate;
+					})
+					.sort(function(a, b) {
+						return a.dueDate > b.dueDate;
+					});
+
+				var basicListActivities = activities
+					.slice(0, 4)
```
CodeBaboon: Does this slice us down to 4 before we do the filter? If so I think we need to do it after.

components/d2l-upcoming-assessments.html
```diff
@@ -87,62 +75,159 @@
...
+					.filter(function(activity) {
+						var activityDueDateTime = new Date(activity.dueDate);
+
+						if (activityDueDateTime - currentDateTime > 0) {
```
CodeBaboon: I think we might be comparing UTC vs local time again.

components/d2l-upcoming-assessments.html
```diff
@@ -87,62 +75,159 @@
...
+				var request = new Request(url, {
+					headers: new Headers({
+						Accept: 'application/vnd.siren+json',
+						Authorization: 'Bearer ' + this.token
+					})
+				});
+
+				return window.d2lfetch
+					.fetch(request)
+					.then(function(response) {
```
CodeBaboon: Should this be using the `_getJson` function above? Otherwise it seems to not have any uses.

components/d2l-upcoming-assessments.html
```diff
@@ -87,62 +75,159 @@
...
+						if (response.ok) {
+							return response.json();
+						}
+						return Promise.reject(response.status);
+					})
+					.then(function(response) {
+						return window.D2L.Hypermedia.Siren.Parse(response);
```
CodeBaboon: Should this be using `_parseEntity`?

demo/index.html
```diff
@@ -5,6 +5,14 @@
 		<meta charset="utf-8">
 		<meta name="viewport" content="width=device-width, minimum-scale=1.0, initial-scale=1, user-scalable=yes">
 		<script src="../../webcomponentsjs/webcomponents-lite.js"></script>
+		<script>
+			'use strict';
+			window.d2lfetch = {
```
CodeBaboon: You could probably just reference the actual `d2lfetch` script?

components/d2l-upcoming-assessments.html
apalaniuk: True - I _sort of_ don't like that format, but that's just a personal thing, so. Conciseness and all.

components/d2l-upcoming-assessments.html
apalaniuk: erm, true, since things in the past would still be considered otherwise.

components/d2l-upcoming-assessments.html
apalaniuk: Hmm - `currentDateTime` is constructed based on local settings (eg. local time and timezone offset), but the underlying implementation is UTC (it represents a single universal moment in time); `activityDueDateTime` is constructed from an explicitly UTC time (eg. `activity.dueDate` is of the form `2017-06-28T12:00:00Z`), so the time range should always be a calculation between two UTC dates, with the activity due date not dependent on the local settings.

Or:
```
// Difference between March 12, 3:00AM in UTC, and March 12, 1:59:59 AM in local time (1 second before DST jump-forward), in milliseconds
new Date('2017-03-12T08:00:00.000Z') - new Date(2017, 2, 12, 1, 59, 59)
1000
```

We could use moment.js instead, but that seems to add a non-trivial amount of bloat to `student-dashboard` (~100KB), which I was hoping to avoid because we don't need _that_ much functionality I don't think. (Granted, adding the calendar in the future _might_ bring moment.js with it, but I don't want to predict that just yet)

CodeBaboon: Awesome ok, seems like this should work as is.

components/d2l-upcoming-assessments.html
apalaniuk: Another casualty from my copypasta from two approaches from ppui. Away it goes!

demo/index.html
apalaniuk: I wanted to explicitly fetching getting external scripts, though in retrospect,`siren-parser` is an implicit external dependency, so I guess not _too_ much is saved in retrospect.

CodeBaboon:
Testing notes _(tested on Chrome, FF, Edge, Android, iOS)_

**Scenarios**

- [x] assignments with no due date do not appear
- [x] assignment due yesterday does not appear
- [x] assignment due earlier today does not appear
- [x] assignment due today shows "Due Today"
- [x] assignment due tomorrow shows "Due Tomorrow"
- [x] assignment due n>1 shows "Due Month Day"
- [x] assignments are ordered by due date
- [x] desktop shows max 4 assignments
- [x] mobile view shows max 3 assignments
- [x] assignments beyond the last day of next week are not displayed ('last day of the week' == saturday, apparently)
- [x] start and end dates do not impact the visibility of assignments
- [x] assignments are aggregated from all courses the user is enrolled in

**Issues**

- empty state message appears while we are fetching the activities and then is replaced with the activity details if they are retrieved
- hidden assignments are returned by the call to https://{tenantid}.activities.api.dev.brightspace.com/users/{userid} but then the subsequent call to https://{tenantid}.assignments.api.dev.brightspace.com/6614/folders/3 fails with 404
- the above error causes the upcoming assignments screen to stay on empty state (should either ignore and show what it gets or else move to error state).
- on user 1, showing 3 assignments, user switcher to user 2, error occurs (dropped connection on phone), error state text appears but the 3 assignments from user 1 are still visible.
- iOS Safari (iPad), assignments do not display

apalaniuk:
The pre-loading issue will be fixed by a PR to parent-portal-ui that shows a skeleton until loaded, but the fixes for the other stuff will probably naturally fix that also. I'll take a look now.

CodeBaboon:
LGTM

CodeBaboon:
Now it looks even better!

CodeBaboon:
I think we are good to merge when Travis gives us the 👍

PR #16, opened 2017-07-04
Pass the activity completion state to the list item view

CodeBaboon:
This is the remaining portion to complete [US88269](https://rally1.rallydev.com/#/128188728532d/detail/userstory/132090331296)

Acceptance Criteria:
- completed things should have a checkmark
- not completed things should not have a checkmark
- the checkmark should look like the one seen here: https://app.zeplin.io/project/58d138705fdb5c5361c515df/screen/58e546640a2e8b6a6e6a27e4

### DIFF
--- a/components/d2l-upcoming-assessments.html
+++ b/components/d2l-upcoming-assessments.html
@@ -224,10 +224,12 @@
 
 					var orgUnitLink = (assignment.getLinkByRel(self.HypermediaRels.organization) || {}).href;
 					var activityUsageLink = assignment.getLinkByRel(self.HypermediaRels.Activities.activityUsage).href;
+					var activityIsComplete = !!assignment.getSubEntityByClass('complete');
 
 					var activityUsageRequest = self._fetchEntity(activityUsageLink)
 						.then(function(activityUsageEntity) {
 							return {
+								activityIsComplete: activityIsComplete,
 								activityUsage: activityUsageEntity,
 								orgUnitLink: orgUnitLink
 							};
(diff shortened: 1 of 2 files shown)


components/d2l-upcoming-assessments.html
```diff
@@ -196,6 +198,7 @@
 
 					var assignmentLink = response.activityUsage.getLinkByRel(self.HypermediaRels.assignment).href;
 					var orgUnitRequest = orgPromiseMap[response.orgUnitLink];
+					var activityIsComplete = response.activityIsComplete;
```
apalaniuk: I think you can use `response.activityIsComplete` below without the intermediate var

CodeBaboon: `response` gets redefined as the incoming function param down on line 212 so I don't think I can.

apalaniuk: Ah, true - though that means I should probably rename the outer scope `response` to be more specific anyway, since I didn't pick a great name for that.

apalaniuk:
I think a couple of things I did previously get in the way of the design a bit (sorry!):

https://github.com/Brightspace/d2l-upcoming-assessments-ui/blob/master/components/d2l-assessments-list-item.html#L44 - `width` and `height` should be 24px, which is the default, so these can be removed entirely I think Don't recall why I manually set it - I'll pretend the icon was smaller some earlier design. 🙄 

https://github.com/Brightspace/d2l-upcoming-assessments-ui/blob/master/components/d2l-assessments-list-item.html#L15 - From a time when margins were calculated differently.. or something. It pushes the completion icon over, so it will be misaligned when this is merged with your activity count PR. Looks like that entire line can also be deleted.

Everything else (including RTL) looks good, and it functions as expected, sooooo..

PR #18, opened 2017-07-05
US87822 - Populate all assignments view

apalaniuk:
https://rally1.rallydev.com/#/128188728532d/detail/userstory/128262149148

Creates the "All Upcoming Activities" view and populates it. There are some minor(ish?) styling issues that I think can come in a second pass to allow all related dependencies/PRs to get in first.

Not yet covered: 
* "Complete"/"Due Today" indicators in overlay - ~will be added when #16 is merged, and it's cleared up whether the design is the same for both mobile and desktop overlay views~ Or, probably in a separate PR
* Ellipsis on instructions overflow: Cross-browser ellipsis truncation in CSS isn't a thing - will evaluate some other solutions/what's acceptable in the future polish PR

Requires: 
https://github.com/Brightspace/parent-portal-ui/pull/239
https://git.dev.d2l/projects/IE/repos/parentportal/pull-requests/124/overview

### DIFF
--- a/components/d2l-all-assessments-list-item.html
+++ b/components/d2l-all-assessments-list-item.html
@@ -60,7 +60,7 @@
 
 			.instructions {
 				@apply --d2l-body-compact-text;
-				font-size: 0.7rem;
+				font-size: 0.8rem;
 				font-weight: 300;
 				margin-top: 12px;
 				max-height: 2.4rem;
--- a/components/d2l-all-assessments-list-item.html
+++ b/components/d2l-all-assessments-list-item.html
@@ -117,7 +117,7 @@
 			}
 
 			return this.localize('dueDateShort', 'dueDate', dueDateString);
-		},
+		}
 
 	});
 	</script>
--- a/components/d2l-all-assessments-list-item.html
+++ b/components/d2l-all-assessments-list-item.html
@@ -68,17 +68,17 @@
 			}
 
 		</style>
-			<div class="row">
-				<d2l-icon class="activity-icon" icon="d2l-tier3:assignments"></d2l-icon>
-				<div class="info-container">
-					<h3 class="assessment-title">[[assessmentItem.name]]</h3>
-					<div class="course-name">[[assessmentItem.courseName]]</div>
-				</div>
+		<div class="row">
+			<d2l-icon class="activity-icon" icon="d2l-tier3:assignments"></d2l-icon>
+			<div class="info-container">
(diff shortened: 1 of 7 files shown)


components/d2l-all-assessments-list.html
```diff
@@ -0,0 +1,118 @@
...
+				display: block;
+				width: 70%;
+				min-width: 300px;
+				margin: auto;
+			}
+
+			.date-header {
+				@apply --d2l-body-compact-text;
+				font-weight: 300;
+				font-size: 0.7rem;
```
CodeBaboon: I think there needs to maybe be a `margin-top: 40px` or something equivalent on all `.date-header`s after the first one.

components/d2l-all-assessments-list.html
```diff
@@ -0,0 +1,118 @@
...
+<link rel="import" href="../src/behaviors/date-behavior.html">
+<link rel="import" href="../src/behaviors/localize-behavior.html">
+<link rel="import" href="d2l-all-assessments-list-item.html">
+
+<dom-module id="d2l-all-assessments-list">
+	<template>
+		<style>
+			:host {
+				display: block;
+				width: 70%;
```
CodeBaboon: This seems to grow a bit too wide at max desktop width 1230px. Mock appears to be about `65%` at that width.

components/d2l-all-assessments-list-item.html
```diff
@@ -0,0 +1,102 @@
...
+
+<dom-module id="d2l-all-assessments-list-item">
+	<template>
+		<style>
+			:host {
+				display: block;
+				margin: 10px auto;
+				background-color: var(--d2l-color-white);
+				border: 1px var(--d2l-color-titanius) solid;
+				border-radius: 6px;
```
CodeBaboon: This appears to be more rounded than the mock, to my eyes `4px;` appears to be a closer match, but if these values are specified somewhere explicitly or are consistent across other areas or something I'm cool with them.

apalaniuk: Nah, I think you're right - these are based on direct comparison + ruler extensions + using similar styling on other components, but are probably off. I'll note all of these and go over them with Bob at the same time for exact values.

components/d2l-all-assessments-list.html
apalaniuk: Yeah, I also noticed that multiple activities on the same date weren't stacking correctly, so I'll lump that together in the next pass.

PR #19, opened 2017-07-07
Creation of a component for displaying upcoming assessment count

CodeBaboon:
[US87825 - Count of all upcoming assignments is displayed in user-selector card](https://rally1.rallydev.com/#/128188728532d/detail/userstory/128262907804)
[US88294 - Display count of overdue/late assignment in user-selector card](https://rally1.rallydev.com/#/128188728532d/detail/userstory/132305570796)
Supports https://github.com/Brightspace/parent-portal-ui/pull/247

Is a ~component~ behavior for displaying upcoming assessment and overdue assessment counts. ~Includes styling of the little box~. Will be initially used on the select-profile user tiles. Potentially could be the thing shown on the upcoming assessments widget chrome as well but until caching (or at least d2lfetch + dedupe) is implemented that would result in some duplicate calls so it's not really worth it at the moment.


components/d2l-upcoming-assessments-count.html
```diff
@@ -0,0 +1,112 @@
...
+							(activities.getSubEntitiesByClass('user-assignment-activity') || [])
+							.map(function(assignment) {
+								return {
+									dueDate: ((assignment.getSubEntityByClass('due-date') || {}).properties || {}).date
+								}
+							})
+							.filter(function(item) {
+								return self.isActivityUpcoming(item);
+							})
+							.length
```
CodeBaboon: At the moment this count will include non-published activities, so it doesn't always match 100% the value on the dashboard widget. I am assured by @ryantmer  that the change to the API so that only published activities are returned is just around the corner.

components/d2l-upcoming-assessments-count.html
```diff
@@ -0,0 +1,112 @@
...
+			_getInfo: function() {
+				var self = this;
+
+				return this._fetchEntity(self.userUrl)
+					.then(function(userEntity) {
+						return self._getUserActivityUsages(userEntity);
+					})
+					.then(function(activities) {
+						self._setTotalCount(
+							(activities.getSubEntitiesByClass('user-assignment-activity') || [])
```
ryantmer: This and the `due-date` could be added to `d2l-hm-constants-behavior` - we're mostly using it for rels granted, but there are some classes in there and there could be more! :D

components/d2l-upcoming-assessments-count.html
```diff
@@ -0,0 +1,112 @@
...
+				if (myActivitiesLink) {
+					return this._fetchEntity(myActivitiesLink);
+				}
+			},
+
+			_fetchEntity: function(url) {
+				var token = this.token;
+
+				return new Promise(function(resolve, reject) {
+					var xhr = new XMLHttpRequest();
```
ryantmer: No `d2l-fetch`? 😮

components/d2l-upcoming-assessments-count.html
apalaniuk: I've forgotten how many times I've forgotten to add those. So.. I'll add those.. promise :wink:

components/d2l-upcoming-assessments-count.html
apalaniuk: I had earlier run into an issue where Safari would randomly (as far as I could tell) throw an exception from native code with zero useful information on use of `d2l-fetch` from within this component, with the same inputs. I stress tested `d2l-fetch` on Safari over the last couple of days and it seems fine.. So I'm going to take another crack at adding it in.

apalaniuk:
Hijacked this slightly to prevent having to do two review/test passes - but since the overdue assignments widget depends on some of the same upcoming assessments code, I turned the "count" behavior into an overall behavior for the component to reduce duplicated code, and moved the logic specific to the profile tile count component from here into PPUI. That said, everything works the same before/after, on children that both have and don't have assignments/overdue assignments, and switching between children in PPUI causes no issue.

PR #21, opened 2017-07-12
Prefer user of instructionsText over instructions

ryantmer:
The former is HTML-less, whereas `instructions` contains the HTML as well.

### DIFF
--- a/behaviors/d2l-upcoming-assessments-behavior.html
+++ b/behaviors/d2l-upcoming-assessments-behavior.html
@@ -73,7 +73,7 @@
 								name: activity.properties.name,
 								courseName: organization.properties.name,
 								dueDate: activity.properties.dueDate,
-								instructions: activity.properties.instructions,
+								instructions: activity.properties.instructionsText || activity.properties.instructions,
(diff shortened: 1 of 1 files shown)


behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -73,7 +73,7 @@
 								name: activity.properties.name,
 								courseName: organization.properties.name,
 								dueDate: activity.properties.dueDate,
-								instructions: activity.properties.instructions,
+								instructions: activity.properties.instructionsText || activity.properties.instructions,
```
apalaniuk: Unlikely, but what if `instructionsText` is an empty string? Will this be `<p></p>`?

(Also, thanks for throwing this in - I have it in a "fix all the things" branch but should have been its own thing, so :+1:)

ryantmer: If there are no instructions, then both would be empty yeah. They come from the same place (text field -> object -> object.Text/object.Html), so ending up `<p></p>` shouldn't ever happen as far as I can tell

PR #22, opened 2017-07-18
Prev next period links to all assignments view

CodeBaboon:
[US88695 - Display week selection dropdown in View All Assignments overlay](https://rally1.rallydev.com/#/89963236876d/detail/userstory/135175762944)

Option 2 was chosen, which is to provide links for moving to other 2 week periods.

TODO:

- [x] Styling. Currently no design direction, have reached out to Bob. **UPDATE**: Styling is to mostly remain as-is. Date formatting has changed to be "July 17 - Aug 1", or if both dates fall in the same month "July 1 - 14".
- [x] Tests.
- [x] Some kind of loading indicator or other indication that stuff is happening when a link is clicked. **UPDATE**: This should be a progress bar.

### DIFF
--- a/components/d2l-upcoming-assessments.html
+++ b/components/d2l-upcoming-assessments.html
@@ -349,11 +349,11 @@
 
 				var startDateString = new window.d2lIntl.DateTimeFormat(this.locale, { format: 'monthDay' }).format(startDateDate);
 				var endDateString;
-				//if (startDateDate.getMonth() === endDateDate.getMonth()) {
-				//	endDateString = endDateDate.getDate();
-				//} else {
-				endDateString = new window.d2lIntl.DateTimeFormat(this.locale, { format: 'monthDay' }).format(endDateDate);
-				//}
+				if (startDateDate && endDateDate && (startDateDate.getMonth() === endDateDate.getMonth())) {
+					endDateString = endDateDate.getDate();
+				} else {
+					endDateString = new window.d2lIntl.DateTimeFormat(this.locale, { format: 'monthDay' }).format(endDateDate);
+				}
 
 				return this.localize('currentPeriod', 'startDate', startDateString, 'endDate', endDateString);
 			},
--- a/components/d2l-upcoming-assessments.html
+++ b/components/d2l-upcoming-assessments.html
@@ -350,7 +350,8 @@
 				var startDateString = new window.d2lIntl.DateTimeFormat(this.locale, { format: 'monthDay' }).format(startDateDate);
 				var endDateString;
 				if (startDateDate && endDateDate && (startDateDate.getMonth() === endDateDate.getMonth())) {
-					endDateString = endDateDate.getDate();
+					//endDateString = endDateDate.getDate();
+					endDateString = new window.d2lIntl.DateTimeFormat(this.locale, { format: 'monthDay' }).format(endDateDate);
 				} else {
 					endDateString = new window.d2lIntl.DateTimeFormat(this.locale, { format: 'monthDay' }).format(endDateDate);
 				}
--- a/components/d2l-upcoming-assessments.html
+++ b/components/d2l-upcoming-assessments.html
@@ -1,6 +1,7 @@
 <link rel="import" href="../../polymer/polymer.html">
 <link rel="import" href="../../d2l-hm-constants-behavior/d2l-hm-constants-behavior.html">
 <link rel="import" href="../../d2l-link/d2l-link.html">
+<link rel="import" href="../../d2l-page-load-progress/d2l-page-load-progress.html">
 <link rel="import" href="../../d2l-simple-overlay/d2l-simple-overlay.html">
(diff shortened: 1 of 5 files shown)


ryantmer:
For item 3, https://github.com/BrightspaceUI/loading-spinner! :) (in case you weren't aware of it, though I suspect you were :P )

components/d2l-upcoming-assessments.html
```diff
@@ -197,21 +212,33 @@
...
+						}
+						self._setLoaded(true);
+					});
+			},
+
+			_getActivityInfo: function(activities) {
+				var self = this;
+
+				self._previousPeriodUrl = (activities.getLinkByRel('https://activities.api.brightspace.com/rels/previous-period') || {}).href;
+				self._nextPeriodUrl = (activities.getLinkByRel('https://activities.api.brightspace.com/rels/next-period') || {}).href;
```
CodeBaboon: I hard-coded these two `rels` instead of adding them to the constants repo for a couple of reasons:

1. Based on conversations with @jopnick it seems that portions of the prev/next navigation within this API may be subject to future change. Those changes may or may not affect these specific rel names, but...
2. This entire UI solution is a stopgap to the real implementation which will involve a calendar picker and arbitrary definition of the time period to retrieve, which will more than likely not require much if any of this existing code.

Given those two things it seemed that rather than risk adding those `rels` to the constants repo and then possibly needing to make a breaking change later on it was probably ok to just hardcode the strings in this mostly throwaway code. Should we end up making use of these `rels` down the line we can easily add them to the repo then.

components/d2l-upcoming-assessments.html
```diff
@@ -263,6 +304,60 @@
...
+			},
+
+			_getFormattedPeriodText: function(startDate, endDate) {
+				var startDateDate = new Date(startDate);
+				var endDateDate = new Date(endDate);
+
+				var startDateString = new window.d2lIntl.DateTimeFormat(this.locale, { format: 'monthDay' }).format(startDateDate);
+				var endDateString;
+				if (startDateDate.getMonth() === endDateDate.getMonth()) {
+					endDateString = endDateDate.getDate();
```
CodeBaboon: Assuming Safari is cool, this should be an ok thing to do because d2l-intl just uses a numeric result for date numbers in the same fashion that `Date.getDate()` will do, so we won't have a mismatch like we would if we used the standard `Intl` formatting alongside `Date.getDate()`.

components/d2l-upcoming-assessments.html
```diff
@@ -97,13 +110,21 @@
...
 			with-backdrop>
-			<div class="view-all-container" slot="header">
-				<d2l-simple-overlay-close-button>
-					<div class="overlay-close-button-content-wrapper">
-						<d2l-icon icon="d2l-tier1:chevron-left"></d2l-icon>
-						<span>[[_closeSimpleOverlayText]]</span>
-					</div>
-				</d2l-simple-overlay-close-button>
+			<div slot="header">
+				<d2l-page-load-progress id="allAssignmentsProgressBar" autostart color="#003b71"></d2l-page-load-progress>
```
CodeBaboon: I should note that I manually applied this color hex because trying to set the property in CSS doesn't seem to actually work with the progress bar.

apalaniuk:
Browsers tested: Win10 + Chrome 59, Firefox 54, macOS 10.12 + Safari 10.1.1, Android 7.1 + Chrome 59, iOS 10.3.2 + Safari (iPad)

- [x] Defaults to current date range on first open
- [x] Shows all activities within this date range** (see below)
- [x] Moving forward shows activities within next period
- [x] Moving backward shows previous period's activities
- [x] When connection is lost and regained, further attempts to move back/forward are successful and data is consistent

A few thoughts:

- We should possibly have it reset to the time period starting today every time the overlay is opened rather than staying at the last viewed date range
- Using `previous-period` and `next-period` are time-based rather than calendar date-based, so on the last date of the period, you'll only get activities that are prior to the cutoff on that day - then the next period will have the previous period's last day as its first day, and also include any remaining activities from that common date. We should see if we can make the whole thing calendar-date based.
- This seems to stray a bit from the "two week period starting from the beginning of this week" model previously talked about, but in the absence of a calendar, doesn't seem like an issue, and the API lends itself well to this workflow right now

Not super critical IMO since it's always clear what date range you're looking at, and the time vs. calendar date issue is for activities two weeks away, but may need to clarify these points in the future.

components/d2l-upcoming-assessments.html
```diff
@@ -263,6 +304,60 @@
...
+					.then(function(activities) {
+						return self._getActivityInfo(activities);
+					})
+					.then(function() {
+						progress.finish();
+					});
+			},
+
+			_getFormattedPeriodText: function(startDate, endDate) {
+				var startDateDate = new Date(startDate);
```
apalaniuk: nit: Maybe overwrite `startDate`/`endDate` since they aren't reused and this is a bit of a tongue twister? Or just call the params `start` and `end`?

PR #27, opened 2017-07-26
US88415 - Use token getting

trevorjonesd2l:
[parent-portal-ui PR](https://github.com/Brightspace/parent-portal-ui/pull/278)


ChrisKraljevicD2L:
Code LGTM, I'll test

test/d2l-upcoming-assessments-behavior.js
```diff
@@ -0,0 +1,91 @@
+/* global describe, it, expect, fixture, beforeEach, sinon */
+
+'use strict';
+
+describe('d2l-upcoming-assessments-behavior', function() {
+	var component,
+		sandbox,
+		token = 'smokin-token',
+		getToken = function() {
```
colinstrancd2l: These assignments _could_ be done in a `before` statement for clarities sake.

test/d2l-upcoming-assessments-behavior.js
```diff
@@ -0,0 +1,91 @@
...
+	});
+
+	describe('fetching enities', function() {
+		it('should not make request if getToken or url is not provided', function(done) {
+			component._makeRequest = sandbox.stub();
+			component._getToken = null;
+
+			setTimeout(function() {
+				component._fetchEntity('url', null, null);
+				expect(component._makeRequest.called).to.be.false;
```
colinstrancd2l: I think tests try to test exactly one thing, which might suggest we split these three `expect`s into individual tests.

test/d2l-upcoming-assessments-behavior.js
```diff
@@ -0,0 +1,91 @@
...
+		component = fixture('d2l-upcoming-assessments-behavior-consumer-fixture');
+		sandbox = sinon.sandbox.create();
+	});
+
+	describe('fetching enities', function() {
+		it('should not make request if getToken or url is not provided', function(done) {
+			component._makeRequest = sandbox.stub();
+			component._getToken = null;
+
+			setTimeout(function() {
```
colinstrancd2l: Polymer does provide a [flush](https://www.polymer-project.org/1.0/docs/tools/tests#test-local-dom) function which I think waits for the dom changes more explicitly than `setTimeout`.

test/d2l-upcoming-assessments-behavior.js
CodeBaboon: Or a datadriven test:

```
[
    { parm1: 'url', parm2: null, parm3: null },
    { parm1: null, parm2: getToken, parm3: null },
    { parm1: null, parm2: null, parm3: url}
].forEach(function(testcase) {
    it(''should not make request if getToken or url is not provided', function() {
        component._fetchEntity(testcase.parm1, testcase.parm2, testcase.parm3);
        expect(component._makeRequest.called).to.be.false;
    });
});
```

(pretend `done()` and `setTimeout` and that stuff are in there somewhere).

test/d2l-upcoming-assessments-behavior.js
```diff
@@ -0,0 +1,91 @@
...
+				expect(component._makeRequest.called).to.be.true;
+				done();
+			});
+		});
+
+		it('should make request when getToken, url and userUrl are provided', function(done) {
+			component._makeRequest = sandbox.stub();
+			component._fetchEntity('url', getToken, 'userUrl');
+			setTimeout(function() {
```
CodeBaboon: for these ones that are making a call to a function and you just want to make sure the promise evaluates you can avoid `setTimeout` and `done()` by doing:

it('should make request when getToken, url and userUrl are provided', function(done) {
    component._makeRequest = sandbox.stub();
    return component._fetchEntity('url', getToken, 'userUrl')
        .then(function() {
            expect(component._makeRequest).to.be.called;
        });
});

CodeBaboon: oh `_makeRequest` also needs to return a resolved promise:

`component._makeRequest = sandbox.stub().returns(Promise.resolve());`

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -157,7 +164,22 @@
...
 
-		_fetchEntity: function(url, token) {
+		_fetchEntity: function(url, getToken, userUrl) {
+			this._getToken = this._getToken || getToken;
+			var tokenUrl = userUrl || url;
+			if (url && typeof this._getToken === 'function' && tokenUrl) {
+				return this._getToken(tokenUrl)
+					.then(function(token) {
+						if (typeof token === 'string') {
+							return this._makeRequest(token, url);
```
ChrisKraljevicD2L: Should `this` be `self` here?

ChrisKraljevicD2L:
Also, these dependency errors in CI are fixed in master now. Rebase ahoy

behaviors/d2l-upcoming-assessments-behavior.html
apalaniuk: In this case, `this` is bound to the function explicitly

CodeBaboon: I think it's bound to `this` on 178

ChrisKraljevicD2L: Goldangit, you are correct, my mistake. I became paranoid after the `this` stuff in the `parent-portal-ui` PR

components/d2l-upcoming-assessments.html
```diff
@@ -199,7 +202,7 @@
 			],
 
 			observers: [
-				'_onApiConfigChanged(userUrl, token)'
+				'_onApiConfigChanged(userUrl)'
```
CodeBaboon: I think we could run into an issue here where the `userUrl` value gets set before the `getToken` property value, causing `_fetchEntity` to not do anything. Then when the `getToken` property is set nothing triggers a new fetch from what I can tell, so we just wouldn't get any data.

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -109,33 +113,36 @@
 			return Promise.all(requests);
 		},
 
-		_getOverdueActivitiesList: function(userUrl, token) {
+		_getOverdueActivitiesList: function(userUrl, getToken) {
+			this._getToken = this._getToken || getToken;
```
CodeBaboon: This feels like strange behavior in here. If we already have a function in `this._getToken` we just ignore what was passed in? Why are we setting a 'class' level variable that affects future calls with a value provided in a function call? As a caller I would not expect that behavior.

components/d2l-upcoming-assessments.html
apalaniuk: I agree - can we just add `getToken` to the observer just as `token` was before? Or, make the API take an object that has all data, and then get rid of debouncing. But the former seems simpler since `getToken` should be able to be added as an observer dependency just as `token` was before.

behaviors/d2l-upcoming-assessments-behavior.html
trevorjonesd2l: Would it be better off just passing `getToken` directly to `fetchEntity` and not setting it at a class level?

CodeBaboon: Probably. Alternatively we could leave the behavior as just wanting the token and let the caller (so the thing that is mixing in this behavior) be the one that cares about token-gettering, so basically keeping the `_makeRequest` logic in here but moving the stuff now in `_fetchEntity` into the component. I'm not entirely certain which is more appropriate with respect to a behavior.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "d2l-upcoming-assessments",
-  "version": "0.18.0",
+  "version": "1.0.0",
```
CodeBaboon: I think `frau-ci` should do this if you add `[increment major]` to the merge notes

ChrisKraljevicD2L:
I'll give this a quick re-test

ChrisKraljevicD2L:
It's gucci! :+1:

PR #29, opened 2017-08-02
US89316 - assessments status badge design update

colinstrancd2l:
[Story](https://rally1.rallydev.com/#/89963236876d/detail/userstory/140528603880)

### DIFF
--- a/components/d2l-all-assessments-list-item.html
+++ b/components/d2l-all-assessments-list-item.html
@@ -52,6 +52,11 @@
 				border-radius: 12px;
 			}
 
+			:host-context([dir="rtl"]) .activity-status-badge {
+				margin-left: 10px;
+				margin-right: 0px;
+			}
+
 			.completion-info {
 				background-color: var(--d2l-color-olivine);
 			}
@@ -62,11 +67,6 @@
 
 			.overdue-info {
 				background-color: var(--d2l-color-cinnabar);
-			}
-
-			:host-context([dir="rtl"]) .completion-info {
-				margin-right: 0px;
-				margin-left: 30px;
 			}
 
 			.activity-info {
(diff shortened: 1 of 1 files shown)


colinstrancd2l:
A quick note, I'm not really sure how `isOverdue`, `isDueToday` and `isCompleted` work but it seems to me that more than one of these badges can appear at a time. If that is the case, should I be reversing the order since it is left-aligned now as opposed to right-aligned?

CodeBaboon:
1) The badges can't appear together on the same item because the service should never return a completed item as overdue, and then because of https://github.com/Brightspace/d2l-upcoming-assessments-ui/blob/master/behaviors/d2l-upcoming-assessments-behavior.html#L100
2) `min-width` is necessary for the case of different badges appearing above/below each other. So on your current page of items you might see an Overdue item and also a Complete on below it. Or maybe a Due Today and also a Complete later. Etc. It would look a bit funny if they are different widths.

colinstrancd2l:
Got news straight from the source that we do not need `min-width` anymore. Just pushed a new commit because that source did mention there was too much room between the assignment title and the meta info

CodeBaboon:
Testing Checklist
Sanity Checks
- [x] ~Verify that unit tests and spec tests have been written, if appropriate~ n/a
- [x] ~Verify that the code is behind a feature flag~ n/a

Acceptance/Functionality
- [x] (UI) Check that the functionality is the same across browsers and devices

Accessibility (UI)
- [x] Verify that workflows can be completed via screen reader/keyboard
- [x] Verify that workflows can be completed when the browser is zoomed to 200%
       ISSUE: Complete badge should be `$d2l-color-olivine-minus-1: #327846 !default;`

Internationalization (UI)
- [x] Verify that lang terms are not hard coded
- [x] Verify that workflows can be completed when RTL

PR #34, opened 2017-08-10
add some more padding between icon and text

Surekh:
Based on feedback from Bob.


Surekh:
Testing notes -
- [x] RTL works as expected. 
- [x] Verified on Safari and Chrome. Change should be independent of browser.

PR #40, opened 2017-08-31
US89512 - Fix isOverdue value

ryantmer:
This should be set to the return value of Array.some. In the previous implementation, the isOverdue value would just end up being "the last comparison value", which could very well be wrong (and often was). Either need to return the isOverdue from within to short-circuit when something truthy is returned, or better yet, just use the returned value of Array.some as isOverdue.

Also simplified the logic to fetch overdue assessments a bit; not sure why there was a Map involved, but since there's only ever one overdue request doesn't need to be.


ryantmer:
Going to merge and release this, then validate the fix prior to releasing parent-portal-ui

PR #42, opened 2017-09-05
US89803: display assessment type icon and name in upcoming assignments widget

margaree:
https://rally1.rallydev.com/#/89963236876/detail/userstory/146719644080

- added icon for quiz and assignment in Upcoming Assessment widget + applicable style
- added assessment type into meta text for assessment in Upcoming Assessment widget
- made "Quiz" and "Assignment" into lang terms

### DIFF
--- a/components/d2l-assessments-list-item.html
+++ b/components/d2l-assessments-list-item.html
@@ -56,8 +56,19 @@
 				font-weight: bolder;
 			}
 
-			.activity-icon, .spacer {
+			.activity-icon-container {
+				width: 24px;
 				margin-right: 20px;
+			}
+
+			:host-context([dir="rtl"]) .activity-icon-container {
+				margin-right: 0px;
+				margin-left: 20px;
+			}
+
+			:host-context([dir="rtl"]) .activity-icon[icon="d2l-tier2:assignments"] {
+				width: 26px;
+				margin-right: -25px;
 			}
 
 			.activity-info {
@@ -68,7 +79,9 @@
 
 		</style>
 			<div class="activity-info">
-				<d2l-icon class="activity-icon" icon="[[_getAssessmentIcon(assessmentItem)]]"></d2l-icon>
+				<div class="activity-icon-container">
+					<d2l-icon class="activity-icon" icon="[[_getAssessmentIcon(assessmentItem)]]"></d2l-icon>
+				</div>
 				<div class="info-container">
 					<h3 class="assessment-title">{{assessmentItem.name}}</h3>
 					<div class="assessment-info">
--- a/components/d2l-assessments-list-item.html
+++ b/components/d2l-assessments-list-item.html
@@ -32,6 +32,8 @@
 				display: flex;
 				flex-direction: row;
 				align-items: center;
+				margin-top: 5px;
+				line-height: 1;
 			}
 
 			.info-container {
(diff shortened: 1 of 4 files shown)


components/d2l-assessments-list-item.html
```diff
@@ -32,6 +32,8 @@
 				display: flex;
 				flex-direction: row;
 				align-items: center;
+				margin-top: 5px;
```
margaree: added this and the line below because else if wrapping is necessary, the line height makes it look weird.

components/d2l-assessments-list-item.html
```diff
@@ -121,6 +140,19 @@ <h3 class="assessment-title">{{assessmentItem.name}}</h3>
...
+			var icon = 'd2l-tier2:';
+			if (assessmentItem.type === 'assignment') return icon + 'assignments';
+			else if (assessmentItem.type === 'quiz') return icon + 'quizzing';
+			else return '';
+		},
+
+		_getAssessmentType: function(assessmentItem) {
+			if (assessmentItem.type === 'assignment') return this.localize('assignment');
+			else if (assessmentItem.type === 'quiz') return this.localize('quiz');
+			else return '';
```
CodeBaboon: super nit, but you don't need the `else`s here because the `if`s do `return`. (same with `_getAssessmentIcon` above).

CodeBaboon:
I can testorize this one

CodeBaboon:
The code review checklist makes me wonder if a test or two could be worth adding? or even just updating [this one](https://github.com/Brightspace/d2l-upcoming-assessments-ui/blob/master/test/components/d2l-assessments-list-item.js#L35) to check for the correct icon perhaps.

CodeBaboon:
Code Review Checklist
- [x] Verify that unit tests and spec tests have been written, if appropriate
- [x] Verify that CI checks are passing
- [x] ~Verify that the code is behind a feature flag~ n/a
- [x] Verify that appropriate documentation has been done (readme, inline, docs.dev, etc.) and makes sense
- [x] Verify to the best of your ability that no security/privacy flaws are present
- [x] Verify that accessibility standards are held to
- [x] Verify that no lang terms are hard coded
- [x] Verify that lang terms have been submitted for translation
- [x] Other review for coding style, functionality, and best practices

Testing Checklist
Sanity Checks
- [x] Verify that unit tests and spec tests have been written, if appropriate
- [x] ~Verify that the code is behind a feature flag~ n/a

Acceptance/Functionality
 ISSUE: On small displays (such as an iPhone) the assignments tier 2 icon will scale the inline-height, however the quiz icon does not, leading to an inconsistent visual as seen here: 
icon-scale

Accessibility (UI)
- [x] Verify that workflows can be completed via screen reader/keyboard
- [x] Verify that workflows can be completed when the browser is zoomed to 200%
- [x] Verify that there are no obvious colour contrast/combination problems

Internationalization (UI)
- [x] Verify that lang terms are not hard coded
  ISSUE: Icons do not align properly with text when RTL
icon-rtl

margaree:
Good catch with the different sized icons. So the reason for the different sizes on smaller screens seems to be related to the amount of information on the line. e.g., when I give a quiz a due date and an end date and then shrink the page, it will resize as well because it has less width. My current thought is to check for screen width and set the width of the icon based on if it's mobile sized. If anyone has any different suggestions let me know.

CodeBaboon:
Is there a lot of value in shrinking the icon at all? It might be simplest to just leave it 24x24 at all times?

ryantmer:
Hmm... I was sure I'd seen this behavior elsewhere in the LMS, where we use slightly smaller icons when the screen goes smaller... but now I can't find where, so maybe I imagined it 🤷‍♂️

CodeBaboon:
This feels like a question for Bob. Was the shrinking by design (and so we should fix it to shrink consistently) or is it unnecessary?

components/d2l-assessments-list-item.html
```diff
@@ -52,19 +56,47 @@
...
+				width: 24px;
+				margin-right: 20px;
+			}
+
+			:host-context([dir="rtl"]) .activity-icon-container {
+				margin-right: 0px;
+				margin-left: 20px;
+			}
+
+			:host-context([dir="rtl"]) .activity-icon[icon="d2l-tier2:assignments"] {
```
margaree: so the assignment icon is mirrored in rtl, so it has a transform applied to it which makes its positioning weird, so this is to compensate for that.

CodeBaboon:
Previous issues from testing are now resolved

PR #51, opened 2017-09-08
Test that assessments and allActivities are set seperately

trevorjonesd2l:
This didn't make it into last PR due to rebase issues. Here it is though, I added a couple tests too

### DIFF
--- a/test/components/d2l-upcoming-assessments.js
+++ b/test/components/d2l-upcoming-assessments.js
@@ -123,6 +123,99 @@ describe('<d2l-upcoming-assessments>', function() {
 					});
 			});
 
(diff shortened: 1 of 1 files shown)


test/components/d2l-upcoming-assessments.js
```diff
@@ -123,6 +123,99 @@ describe('<d2l-upcoming-assessments>', function() {
 					});
 			});
 
+			it('should update the allActivies with the activities in the period', function() {
+				var userUsage = {};
+				userUsage.getLinkByRel = sandbox.stub().returns();
```
ryantmer: Eh, nitpicky, but pretty sure you can just `sandbox.stub()` here (as well as a few times below)

PR #52, opened 2017-09-11
DE27543 Add placeholders for all langterms

ryantmer:
Without these, nothing is rendered when these are required. With these, we at least get the english text.

### DIFF
--- a/lang/ar.html
+++ b/lang/ar.html
@@ -13,13 +13,23 @@
 	*/
 	window.D2L.UpcomingAssessments.LocalizeBehavior.LangArBehavior = {
 		ar: {
+			'activityComplete': 'Complete',
+			'activityDueToday': 'Due Today',
+			'activityEnded': 'Closed',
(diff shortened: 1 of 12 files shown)


Surekh:
@CodeBaboon - Usually @trevorjonesd2l sends a snapshot of all terms and receives a snapshot. This change should not affect anything. UI is actual usable now even if not translated.

CodeBaboon:
How do they use the snapshot to determine what to translate? I like that we are making the UI usable without translations but I just want to ensure we aren't causing problems downstream.

ryantmer:
I don't think we are; I think in general we should be adding an English placeholder whenever we add a new langterm (rather than just adding the English term as we had previously been doing).

CodeBaboon:
Sounds good to me then!

Surekh:
@CodeBaboon - I am not sure what tool localization team uses. But the recommendation is to send snapshot and not diff.

PR #53, opened 2017-09-11
Cover all possible test cases

ryantmer:
The previous test cases only really covered the ideal cases/happy-paths, which was good, but let a few bugs through. This now covers all possible combinations of the variables that we use to determine which badge to show on an activity.

### DIFF
--- a/test/components/d2l-all-assessments-list-item.js
+++ b/test/components/d2l-all-assessments-list-item.js
@@ -5,91 +5,13 @@
 describe('<d2l-all-assessments-list-item>', function() {
 	var element;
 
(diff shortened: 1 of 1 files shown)


ryantmer:
Note that this is just something I didn't finish last sprint that I said I'd do, so it's not very inspirational and I apologize for that 😜

ryantmer:
Mostly I just like to artificially inflate the number of tests so I can lie to myself about test coverage. It makes me feel good.

CodeBaboon:
Now I'm inspired by your simultaneous honesty and dishonesty.

PR #55, opened 2017-09-12
DE27561 _getInfo now calls custom date range action for 2 weeks of work

ChrisKraljevicD2L:
[Rally Defect](https://rally1.rallydev.com/#/89963236876d/detail/defect/152462445840)

This will fix the regression where upcoming work displays only one week of work, instead of two.

### DIFF
--- a/components/d2l-upcoming-assessments.html
+++ b/components/d2l-upcoming-assessments.html
@@ -273,6 +273,10 @@
 
 								var basicListActivities = upcomingActivities.slice(0, 4);
 								self.set('_assessments', basicListActivities);
+							})
+							.catch(function() {
+								self._showError = true;
+								self._userName = null;
 							});
 					});
 			},
--- a/components/d2l-upcoming-assessments.html
+++ b/components/d2l-upcoming-assessments.html
@@ -352,6 +352,7 @@
 					.catch(function() {
 						self._showError = true;
 						self._userName = null;
+						return Promise.reject(new Error('Failed to fetch user activities'));
 					});
 			},
 
--- a/components/d2l-upcoming-assessments.html
+++ b/components/d2l-upcoming-assessments.html
@@ -260,7 +260,10 @@
 					.then(function(userEntity) {
 						self._userName = (userEntity.getSubEntityByRel(self.HypermediaRels.firstName) || { properties: {} }).properties.name;
 						var myActivitiesLink = (userEntity.getLinkByRel(self.HypermediaRels.Activities.myActivities) || {}).href;
-						return self._loadActivitiesForPeriod(myActivitiesLink)
+						return self._getCustomRangeAction(myActivitiesLink)
+							.then(function(customActivitiesLink) {
+								return self._loadActivitiesForPeriod(customActivitiesLink);
+							})
 							.then(function(activities) {
 								var upcomingActivities = activities
 									.filter(function(activity) {
@@ -329,6 +332,58 @@
 				if ( e.code === 'Space' || e.code === 'Enter' ) {
 					return this._goNext(e);
 				}
+			},
+
+			_getCustomRangeAction: function(activitiesUrl) {
+				if (!activitiesUrl) {
+					return Promise.resolve();
+				}
+
+				var self = this;
+
+				return this._fetchEntity(activitiesUrl, this.getToken, this.userUrl)
+					.then(function(activities) {
+						var parameters = self._getCustomDateRangeParameters();
+
+						var action = (activities.getActionByName(self.HypermediaActions.activities.selectCustomDateRange) || {});
+
+						return self._createActionUrl(action, parameters);
+					})
+					.catch(function() {
+						self._showError = true;
+						self._userName = null;
+					});
+			},
+
+			_getCustomDateRangeParameters: function() {
+				var currentTime = new Date();
+				var start = currentTime.toISOString();
+				var twoWeeksFromNow = new Date(currentTime.setDate(currentTime.getDate() + 14));
+				var end = twoWeeksFromNow.toISOString();
+
+				return {
+					start: start,
+					end: end
+				};
+			},
+
+			_createActionUrl: function(action, parameters) {
+				var query = {};
+				if (action.fields) {
+					action.fields.forEach(function(field) {
+						if (parameters.hasOwnProperty(field.name)) {
+							query[field.name] = parameters[field.name];
+						} else {
+							query[field.name] = field.value;
+						}
+					});
+				}
+
+				var queryString = Object.keys(query).map(function(key) {
+					return key + '=' + query[key];
+				}).join('&');
+
+				return queryString ? action.href + '?' + queryString : action.href;
 			},
 
 			_loadActivitiesForPeriod: function(periodUrl) {
--- a/components/d2l-upcoming-assessments.html
+++ b/components/d2l-upcoming-assessments.html
@@ -315,7 +315,11 @@
 			},
 
 			_goPrev: function() {
-				return this._loadActivitiesForPeriod(this._previousPeriodUrl);
+				return this._loadActivitiesForPeriod(this._previousPeriodUrl)
+					.catch(function() {
+						self._showError = true;
+						self._userName = null;
+					});
 			},
 
 			_keypressGoPrev: function(e) {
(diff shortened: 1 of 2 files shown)


components/d2l-upcoming-assessments.html
```diff
@@ -331,6 +334,58 @@
...
+						self._showError = true;
+						self._userName = null;
+					});
+			},
+
+			_getCustomDateRangeParameters: function() {
+				var currentTime = new Date();
+				var start = currentTime.toISOString();
+				var twoWeeksFromNow = new Date(currentTime.setDate(currentTime.getDate() + 14));
+				var end = twoWeeksFromNow.toISOString();
```
CodeBaboon: Since we are doing custom periods now, are we sure we want to start at `now`, rather than say, 'the start of today' or probably even better 'the start of this week'? Or is something like that scheduled as a story that will follow up in a proper release rather than as a defect fix?

test/components/d2l-upcoming-assessments.js
```diff
@@ -125,6 +125,24 @@ describe('<d2l-upcoming-assessments>', function() {
...
+
+			it('calls _fetchEntity for the provided url', function() {
+				element._fetchEntity = sandbox.stub().returns(Promise.resolve(activities));
+				return element._getCustomRangeAction(nextPeriodUrl)
+					.then(function() {
+						expect(element._fetchEntity).to.have.been.calledWith(nextPeriodUrl);
+					});
+			});
+		});
+
```
CodeBaboon: It would be nice to see some tests around the additional functionality (getting the custom parameters, building the action url, etc). Maybe as a follow up at least. Also, something like `_createActionUrl` seems like an awesome helper kind of function that could get re-used all over the place.

ChrisKraljevicD2L: I stole `_createActionUrl` from [here](https://github.com/Brightspace/d2l-my-courses-ui/blob/9a2c7a0779b4c4f510b19ab846b5b176e4974bef/d2l-utility-behavior.html#L16) so it is indeed something good to have re-used

ChrisKraljevicD2L: And yes, follow-up for extra testing stuff would be bueno. I followed what had been written for `_loadActivitiesForPeriod` but I agree that the guts of those two should be tested some more. (I gotta figure out how to mock a siren entity first)

components/d2l-upcoming-assessments.html
ChrisKraljevicD2L: That is scheduled in [this story](https://rally1.rallydev.com/#/89963236876d/detail/userstory/137138049580) somewhat poorly defined but the idea is to have the period start at last Sunday 12:00 AM and go to next Saturday 11:59 PM (~2 weeks).

components/d2l-upcoming-assessments.html
```diff
@@ -331,6 +334,58 @@
 				}
 			},
 
+			_getCustomRangeAction: function(activitiesUrl) {
+				if (!activitiesUrl) {
+					return Promise.resolve();
```
apalaniuk: Shouldn't this reject instead?

ChrisKraljevicD2L: I figured it should, however this pattern is used by [`_loadActivitiesForPeriod`](https://github.com/Brightspace/d2l-upcoming-assessments-ui/blob/31b3cc4f3a3b356a4e3cbb46afd6f27d265ac7ad/components/d2l-upcoming-assessments.html#L390) too. Both of their tests blow up if this isn't present too (which should probably mean their tests should be changed, and I'd gladly do that after inspiration)

components/d2l-upcoming-assessments.html
```diff
@@ -331,6 +334,58 @@
...
+
+				return this._fetchEntity(activitiesUrl, this.getToken, this.userUrl)
+					.then(function(activities) {
+						var parameters = self._getCustomDateRangeParameters();
+
+						var action = (activities.getActionByName(self.HypermediaActions.activities.selectCustomDateRange) || {});
+
+						return self._createActionUrl(action, parameters);
+					})
+					.catch(function() {
```
apalaniuk: I think the subsequent handlers (eg. https://github.com/Brightspace/d2l-upcoming-assessments-ui/pull/55/files#diff-62ae692de3e59ad4f90577ba9e4a3a3dR264) will still be called after this catch - is this intended? Seems like this could lead to console errors and such due to things being undefined.

ChrisKraljevicD2L: Not intended, should probably stop in its tracks. I'll push an update shortly

components/d2l-upcoming-assessments.html
apalaniuk: Hrm. I think any case where that causes either method to return a resolved promise will cause an exception here: https://github.com/Brightspace/d2l-upcoming-assessments-ui/pull/55/files#diff-62ae692de3e59ad4f90577ba9e4a3a3dL266

components/d2l-upcoming-assessments.html
apalaniuk: I like that you put it in here, because I think this wasn't handled before - maybe it could be put on the topmost chain? Seems like a reasonable thing to do in every case.

test/components/d2l-upcoming-assessments.js
```diff
@@ -125,6 +130,29 @@ describe('<d2l-upcoming-assessments>', function() {
 
 		});
 
+		describe('_getCustomRangeAction', function() {
+			it('does nothing if the provided url was not set', function() {
+				element._fetchEntity = sandbox.stub();
+				return element._getCustomRangeAction()
+					.then(function() {
+						return Promise.reject('Expected _getCustomRangeAction to reject');
```
apalaniuk: Just a note that [chai-as-promised](https://www.npmjs.com/package/chai-as-promised) allows you to write more expressive assertions about promises. Not expecting this to be changed though.

ryantmer: Are we able to use arbitrary `chai` packages though? I thought that, since we were using it via the polymer-cli, we didn't have such freedoms :(

components/d2l-upcoming-assessments.html
```diff
@@ -312,7 +319,11 @@
 			},
 
 			_goPrev: function() {
-				return this._loadActivitiesForPeriod(this._previousPeriodUrl);
+				return this._loadActivitiesForPeriod(this._previousPeriodUrl)
+					.catch(function() {
+						self._showError = true;
```
apalaniuk: Do we want to do this here? I think the effect will be that the main widget will show an error rather than upcoming activities, but its filtered list of activities is in the (now poorly named) `_assessments` property, which isn't touched by the custom date range logic - so we should still be able to show it. 

Maybe it's not unreasonable for now, in the absence of the overlay itself having its own error message (which it probably should?)

ChrisKraljevicD2L: That makes sense to not nuke activities that came in just fine, I was mostly concerned about not `catch`ing a rejected promise here, but it should be alright to let it fly into the console.

apalaniuk: Yeah, I think this is okay until we have an error message in the overlay itself - then setting that error message would be what's set here.

test/components/d2l-upcoming-assessments.js
apalaniuk: Bah, true, would have to browserify it first. Wonder if it's worth making a repo for that so we can just import a script once and have that functionality available.. 

But yeah, anyway, apparently wouldn't have been free, so ignore me :P

ryantmer: It would be pretty legit; it's kinda butts that we're limited in which packages for `chai`/`sinon`/etc. we can leverage, but such is life

tiffjiang:
Testing Checklist
Sanity Checks
- [x] Verify that unit tests and spec tests have been written, if appropriate
- [x] Verify that the code is behind a feature flag

Acceptance/Functionality
- [x] (UI) Check that the functionality is the same across browsers and devices

Accessibility and Internationalization checks omitted

Notes:
- The dates are all correct (2 weeks from today)
- The 'Previous' and 'Next' functionalities sometimes (randomly) does not load, especially on Safari & on all browsers on iOS and Android. If you try to go back after encountering this, other errors such as the 'View all work' link disappearing or not being able to scroll/click on the list of children profiles occurs. @apalaniuk

PR #57, opened 2017-09-18
fix font size to make it consistent with Portfolio widget

Surekh:
Thought this was a quick fix, so made it.

[DE27566](https://rally1.rallydev.com/#/89963236876d/detail/defect/152538573176)

### DIFF
--- a/components/d2l-upcoming-assessments.html
+++ b/components/d2l-upcoming-assessments.html
@@ -56,12 +56,12 @@
 
 			.error-message,
 			.no-upcoming-assessments,
-			.view-all-assignments-link > h3,
 			.view-all-container {
 				@apply --d2l-body-compact-text;
 			}
 
 			.view-all-assignments-link > h3 {
+				font-size: 0.7rem;
 				margin-bottom: 0;
 			}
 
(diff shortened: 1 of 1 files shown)


components/d2l-upcoming-assessments.html
```diff
@@ -56,12 +56,12 @@
 
 			.error-message,
 			.no-upcoming-assessments,
-			.view-all-assignments-link > h3,
 			.view-all-container {
 				@apply --d2l-body-compact-text;
 			}
 
 			.view-all-assignments-link > h3 {
+				font-size: 0.7rem;
```
CodeBaboon: Unless I'm mistaken (super possible) I think this change is going to modify additional properties due to the loss of `--d2l-body-compact-text` that we don't want to change:

- `font-weight` (from `400` to `700`)
- `line-height` (from `1.2` to `1.5`)

Surekh: Good point. Didn't noticed it in the demo app. But, I will check later tonight.

Surekh: @CodeBaboon - Portfolio widget and this widget have line-height: 28px/1.4 rem. So not modifying that.

Surekh:
same Looks similar so merging.

PR #58, opened 2017-09-20
DE27639 - Reset error state whenever we load new info

ryantmer:
Previously, if the widget errored out for one child, and then the parent switched children, the widget would remain in an error state. Simple fix, reset to be non-errored before we fetch a different child's data.

Also replaced the commented-out tests with some new, less-commented-out ones.

This isn't a full fix for DE27639, but it's part of what we're seeing in that.


ryantmer:
Ping for review/merge? :D

PR #60, opened 2017-09-29
Us88845 week selection

margaree:
https://rally1.rallydev.com/#/89963236876/detail/userstory/137138049580

This story was separated, so for this part
- Remove "previous" and "next"
- Add calendar "dropdown" (much of the date dropdown is from https://github.com/BrightspaceUI/date-picker/blob/master/d2l-date-picker.html though some customizations were made)
- When a date is selected in the dropdown, show the period of time from the previous Sunday (or midnight of that day if Sunday is selected) for the next 2 weeks, so until 11:59 PM of the Saturday ~2 weeks later

Note that the pointer above the calendar isn't working correct on RTL, though everything else should be working RTL. I'll fix that Monday.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -5,7 +5,7 @@
   "dependencies": {
     "app-localize-behavior": "PolymerElements/app-localize-behavior#^1.0.2",
     "d2l-colors": "^2.2.3",
-    "d2l-date-picker": "^0.0.12",
+    "d2l-date-picker": "~0.0.12",
     "d2l-hm-constants-behavior": "Brightspace/d2l-hm-constants-behavior#^2.4.0",
     "d2l-icons": "^3.1.2",
     "d2l-intl": "^0.4.1",
--- a/bower.json
+++ b/bower.json
@@ -14,13 +14,18 @@
     "d2l-simple-overlay": "Brightspace/simple-overlay#1.1.0",
     "d2l-typography": "^5.3.0",
     "fetch": "^2.0.3",
+    "iron-a11y-keys": "^1.0.9",
     "iron-icon": "^1.0.12",
     "polymer": "Polymer/polymer#^1.9.2",
-    "vaadin-date-picker": "^2.0.2"
+    "vaadin-date-picker": "~1.2.3"
   },
   "devDependencies": {
     "webcomponentsjs": "webcomponents/webcomponentsjs#^0.7.0",
     "d2l-demo-template": "~0.0.11",
     "web-component-tester": "^6.1.3"
+  },
+  "resolutions": {
+    "d2l-polymer-behaviors": "^1.0.0",
+    "d2l-offscreen": "^2.2.5"
   }
 }
--- a/bower.json
+++ b/bower.json
@@ -5,6 +5,7 @@
   "dependencies": {
     "app-localize-behavior": "PolymerElements/app-localize-behavior#^1.0.2",
     "d2l-colors": "^2.2.3",
+    "d2l-date-picker": "^0.0.12",
     "d2l-hm-constants-behavior": "Brightspace/d2l-hm-constants-behavior#^2.4.0",
     "d2l-icons": "^3.1.2",
     "d2l-intl": "^0.4.1",
--- a/bower.json
+++ b/bower.json
@@ -14,7 +14,8 @@
     "d2l-typography": "^5.3.0",
     "fetch": "^2.0.3",
     "iron-icon": "^1.0.12",
-    "polymer": "Polymer/polymer#^1.9.2"
+    "polymer": "Polymer/polymer#^1.9.2",
+    "vaadin-date-picker": "^2.0.2"
   },
   "devDependencies": {
     "webcomponentsjs": "webcomponents/webcomponentsjs#^0.7.0",
--- a/components/d2l-date-dropdown.html
+++ b/components/d2l-date-dropdown.html
@@ -29,7 +29,7 @@
 					--vaadin-date-picker-overlay: {
 						font-family: inherit;
 						max-height: 355px;
-						margin-top: -10px;
+						margin-top: 0;
 						border-radius: 5px;
 						border: 1px solid var(--d2l-color-mica);
 						opacity: 0;
@@ -80,7 +80,7 @@
 					z-index: 10000;
 					display: inline-block;
 					opacity: 0;
-					top: -20px;
+					top: -10px;
 				}
 
 				:host-context([dir="rtl"]) .d2l-dropdown-content-pointer {
@@ -176,7 +176,7 @@
 				var self = this;
 				this.addEventListener('iron-overlay-opened', function() {
 					self._opened = true;
-					self.$$('vaadin-date-picker-light').customStyle['--vaadin-date-picker-overlay'] = 'font-family: inherit; max-height: 355px; margin-top: -10px; border-radius: 5px; border: 1px solid var(--d2l-color-mica); transition: all 200ms cubic-bezier(.75, 0, 1, 1); opacity:1; transform: translateY(20px);';
+					self.$$('vaadin-date-picker-light').customStyle['--vaadin-date-picker-overlay'] = 'font-family: inherit; max-height: 355px; margin-top: 0px; border-radius: 5px; border: 1px solid var(--d2l-color-mica); transition: all 300ms ease; opacity:1; transform: translateY(10px);';
 					self.$$('vaadin-date-picker-light').updateStyles();
 				});
 
@@ -186,7 +186,7 @@
 
 				this.addEventListener('iron-overlay-closed', function() {
 					self._opened = false;
-					self.$$('vaadin-date-picker-light').customStyle['--vaadin-date-picker-overlay'] = 'font-family: inherit; max-height: 355px; margin-top: -10px; border-radius: 5px; border: 1px solid var(--d2l-color-mica); opacity:0;';
+					self.$$('vaadin-date-picker-light').customStyle['--vaadin-date-picker-overlay'] = 'font-family: inherit; max-height: 355px; margin-top: 0px; border-radius: 5px; border: 1px solid var(--d2l-color-mica); opacity:0;';
 					self.$$('vaadin-date-picker-light').updateStyles();
 				});
 			},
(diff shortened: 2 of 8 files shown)


CodeBaboon:
Code Review Checklist
- [ ] Verify that unit tests and spec tests have been written, if appropriate
- [ ] Verify that CI checks are passing
- [ ] Verify that the code is behind a feature flag
- [ ] Verify that appropriate documentation has been done (readme, inline, docs.dev, etc.) and makes sense
- [ ] Verify to the best of your ability that no security/privacy flaws are present
- [ ] Verify that accessibility standards are held to
- [ ] Verify that no lang terms are hard coded
- [ ] Verify that lang terms have been submitted for translation
- [ ] Other review for coding style, functionality, and best practices

Testing Checklist
Sanity Checks
- [ ] Verify that unit tests and spec tests have been written, if appropriate
- [ ] Verify that unit tests, spec tests, and linting are all passing
- [ ] Verify that no code review comments are outstanding
- [ ] Verify that the code is behind a feature flag

Acceptance/Functionality
- [ ] Check that the acceptance criteria of the story has been fulfilled
- [ ] Check that related functionality has not been severely affected (smoke test)

Accessibility (UI)
- [ ] Verify that workflows can be completed via screen reader/keyboard
- [ ] Verify that workflows can be completed when the browser is zoomed to 200%

Internationalization (UI)
- [ ] Verify that lang terms are not hard coded

CodeBaboon:
One issue I'm noticing in testing is that the displayed periods appear to run sun-sun, rather than sun-sat as described in the details above.

sunday-not-saturday

CodeBaboon:
With respect to keyboard support, the calendar control itself works great but I can't actually get into it with the keyboard, I have to click the chevron with my mouse because I can't tab to or otherwise select it.

bower.json
```diff
@@ -5,6 +5,7 @@
   "dependencies": {
     "app-localize-behavior": "PolymerElements/app-localize-behavior#^1.0.2",
     "d2l-colors": "^2.2.3",
+    "d2l-date-picker": "~0.0.12",
```
CodeBaboon: It seems we only use the `d2l-date-picker` for the `localize-behavior` but instead use the `vaadin-date-picker` as the actual date picker, is that correct? Are there things lacking from the `d2l-date-picker`? I'm not familiar with it, maybe it just doesn't do what I think it seems like it should? Mostly just curious how this is all set up.

components/d2l-date-dropdown.html
```diff
@@ -0,0 +1,299 @@
...
+				:host {
+					display: block;
+				}
+
+				vaadin-date-picker-light {
+					height: 28px;
+					--primary-color: var(--d2l-color-celestine);
+					--light-primary-color: var(--d2l-color-celestine-plus-1);
+					--dark-theme-background-color: var(--d2l-color-ferrite);
+					--dark-theme-secondary-color: #ffffff;
```
CodeBaboon: probably should be `var(--d2l-color-white);`

components/d2l-date-dropdown.html
```diff
@@ -0,0 +1,299 @@
...
+				.d2l-dropdown-content-pointer {
+					position: relative;
+					clip: rect(-5px, 21px, 8px, -3px);
+					left: calc(50% - 7px);
+					z-index: 10000;
+					display: none;
+				}
+
+				.d2l-dropdown-content-pointer > div {
+					background-color: #ffffff;
```
CodeBaboon: probably should be `var(--d2l-color-white);`

bower.json
margaree: That's correct. So `d2l-date-picker` uses the `vaadin-date-picker-light` as well but instead of using a link to be clicked to display the calendar, as desired here, it has an input text box. We also will want it to behave like a `d2l-dropdown`, which `d2l-date-picker` does not. What we want in this case is a little too different from `d2l-date-picker` to be able to just use it, and it seemed like a lot of customizations to make to it for what is likely a one-off usage.

components/d2l-date-dropdown.html
```diff
@@ -0,0 +1,317 @@
...
+			ready: function() {
+				this._target = this.$$('a');
+				var mql = window.matchMedia('(max-width: 420px)');
+				mql.addListener(this._mobileBreakpointListener.bind(this));
+
+				var self = this;
+				this.addEventListener('iron-overlay-opened', function() {
+					self._opened = true;
+					self.$$('vaadin-date-picker-light').customStyle['--vaadin-date-picker-overlay'] = 'font-family: inherit; max-height: 355px; margin-top: 0px; border-radius: 5px; border: 1px solid var(--d2l-color-mica); transition: all 300ms ease; opacity:1; transform: translateY(10px);';
+					self.$$('vaadin-date-picker-light').updateStyles();
```
CodeBaboon: I think these styles could be applied by adding/removing classes.

components/d2l-date-dropdown.html
```diff
@@ -0,0 +1,317 @@
...
+				});
+
+				this.addEventListener('iron-overlay-canceled', function(e) {
+					e.stopPropagation();
+				});
+
+				this.addEventListener('iron-overlay-closed', function() {
+					self._opened = false;
+					self.$$('vaadin-date-picker-light').customStyle['--vaadin-date-picker-overlay'] = 'font-family: inherit; max-height: 355px; margin-top: 0px; border-radius: 5px; border: 1px solid var(--d2l-color-mica); opacity:0;';
+					self.$$('vaadin-date-picker-light').updateStyles();
```
CodeBaboon: Same comment as above with respect to CSS classes.

components/d2l-date-dropdown.html
```diff
@@ -0,0 +1,317 @@
...
+				var pointer = this.$$('.d2l-dropdown-content-pointer');
+
+				if (opened && !smallScreen) {
+					pointer.style.transition = 'all 300ms ease';
+					pointer.style.transform = 'translateY(10px)';
+					pointer.style.opacity = 1;
+				} else {
+					pointer.style.transition = '';
+					pointer.style.transform = '';
+					pointer.style.opacity = 0;
```
CodeBaboon: I believe this could also be moved into CSS classes

components/d2l-date-dropdown.html
```diff
@@ -0,0 +1,317 @@
...
+			},
+			observers: [
+				'_localeChanged( locale )',
+				'_updateDatePickeri18n( locale, localize, dateFormat, firstDayOfWeek )',
+				'_showPointer( _smallScreen, _opened )'
+			],
+			ready: function() {
+				this._target = this.$$('a');
+				var mql = window.matchMedia('(max-width: 420px)');
+				mql.addListener(this._mobileBreakpointListener.bind(this));
```
jstefaniuk-d2l: I think this can just be achieved with breakpoints in the css

components/d2l-date-dropdown.html
jstefaniuk-d2l: As a general rule, try to avoid `xxx.style.yyy = zzz` when possible (I find that in frameworks like polymer I _almost_ never have to use them for most cases)

components/d2l-date-dropdown.html
```diff
@@ -0,0 +1,317 @@
...
+
+				var self = this;
+				this.addEventListener('iron-overlay-opened', function() {
+					self._opened = true;
+					self.$$('vaadin-date-picker-light').customStyle['--vaadin-date-picker-overlay'] = 'font-family: inherit; max-height: 355px; margin-top: 0px; border-radius: 5px; border: 1px solid var(--d2l-color-mica); transition: all 300ms ease; opacity:1; transform: translateY(10px);';
```
jstefaniuk-d2l: How many things do you need transitioned?

I've shot myself in the foot with `transition: all` a few times before

components/d2l-date-dropdown.html
jstefaniuk-d2l: And I think if they're classes pp-ui will also give us some autoprefixer goodness which is nice (If I'm right)

CodeBaboon: Indeed. https://github.com/Brightspace/parent-portal-ui/blob/master/gulpfile.js#L140

margaree: I don't think it's actually possible to update the styles by adding/removing classes in the usual way in this case because it's done through a mixin (see https://www.polymer-project.org/1.0/docs/devguide/styling "Dynamism limitations"). From one of the examples, it looks like it would be potentially possible to move the style into the CSS to make it more clear, though the `updateStyles` would still be needed.

components/d2l-date-dropdown.html
```diff
@@ -0,0 +1,312 @@
...
+				}
+			</style>
+		</custom-style>
+
+		<vaadin-date-picker-light value="{{value}}" on-value-changed="handleValueChanged" attr-for-value="value">
+			<iron-a11y-keys target="[[_target]]" keys="enter" on-keys-pressed="_onEnter"></iron-a11y-keys>
+			<iron-a11y-keys target="[[_target]]" keys="up down" on-keys-pressed="_onUpDown"></iron-a11y-keys>
+			<div class="input">
+				<a
+					href="javascript:void(0);"
```
CodeBaboon: I think you can just not include the `href` attribute.

CodeBaboon:
Tried it out again, if I have a period from Oct 1-14, and I click on any date in the second week (so 8-14), it changes the period to be Oct 8-21, which I don't believe was how it functioned in earlier. Is that an intentional change?

components/d2l-date-dropdown.html
margaree: this allows it to be tabbed to via keyboard

margaree:
That was an intentional change and does follow the design in the video in the story

CodeBaboon:
There appears to still be an issue around the end date of the chosen period. In testing, if I create an assignment that is due at 5:00pm on October 14th, it shows up nicely in the upcoming work widget

due-oct14

But when I click into the View All Work link it is not there:

due-oct14-missing

Any date I pick on the first week of the period (Oct 1-7) will not display the item, but if I select a date on the 2nd week (thus changing the period to Oct 8-21) it will appear:

due-oct14-foundyou

It correctly will not display if I choose a date beyond the 14th (such as the 15th). As further info this does not appear to be an issue with due dates even 1 day prior (Oct 13 11:30pm in this case).

components/d2l-date-dropdown.html
```diff
@@ -0,0 +1,312 @@
...
+						} catch (exception) {
+							return null;
+						}
+					}
+				};
+				var datePickerLocale = this._merge(datePicker.i18n, localeOverrides);
+				datePicker.i18n = {}; // reassign to have Polymer refresh
+				datePicker.i18n = datePickerLocale;
+			},
+			_merge: function(obj1, obj2) {
```
jstefaniuk-d2l: If we don't need to support IE11 (I forget if we do), we can maybe use `Object.assign` here

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign

```
return Object.assign(obj1 || {}, obj2 || {});
```

jstefaniuk-d2l: Er, nvm, i don't think Object.assign does a deep-merge like this

PR #62, opened 2017-10-06
Simplify eslint configs

apalaniuk:
Because this may be used as a reference in other repos (eg. https://github.com/Brightspace/d2l-recent-grades-ui/pull/11), and because it was kind of crazy and inaccurate anyway.

### DIFF
--- a/.eslintrc.yaml
+++ b/.eslintrc.yaml
@@ -1,17 +1,7 @@
 extends:
-  - brightspace
-env:
-  browser: true
-  node: true
-plugins:
-  - html
+  brightspace/polymer-config
 rules:
-  no-var: off
+  object-shorthand: [2, "never"]
 globals:
-  a11ySuite: true
-  expect: true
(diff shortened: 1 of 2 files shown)


.eslintrc.yaml
```diff
@@ -1,17 +1,7 @@
...
-  - brightspace
-env:
-  browser: true
-  node: true
-plugins:
-  - html
+  brightspace/polymer-config
 rules:
-  no-var: off
+  object-shorthand: [2, "never"]
```
apalaniuk: `parent-portal-ui`'s build process will happily choke on minifying files using es6 object shorthand without returning any useful information, which is a pain to debug, as debugging the component directly eg. in the demo page won't reveal anything useful (as it's browser supported).

Because the repo now uses es5 syntax by default, not strictly necessary in the current setup as linting _will_ result in syntax errors, but left on in case es6 is flipped back on.

PR #63, opened 2017-10-17
refactor to use d2-fetch-siren-entity-behavior and also move to using…

CodeBaboon:
… Chrome as the local test browser. This was done because Selenium is having issues with current FF for many devs.


ChrisKraljevicD2L:
Yeah totes, sorry for the late reply

PR #64, opened 2017-10-18
US89797 - Allow course name to scale and wrap to next line

tiffjiangd2l:
[Related parent-portal-ui PR](https://github.com/Brightspace/parent-portal-ui/pull/376)

The upcoming assessments widget now looks something like this for a long course name:
capture

This change also required changes to parent-portal-ui in order for the widget heights to scale properly to the longest widget.

### DIFF
--- a/components/d2l-assessments-list-item.html
+++ b/components/d2l-assessments-list-item.html
@@ -82,11 +82,7 @@
 				overflow-wrap: break-word;
 			}
 
-			.assessment-type {
-				display: inline;
-			}
-
-			.date {
+			.date, .assessment-type {
 				display: inline;
 			}
 		</style>
--- a/components/d2l-assessments-list-item.html
+++ b/components/d2l-assessments-list-item.html
@@ -37,6 +37,7 @@
 
 			.info-container {
 				display: block;
+				width: calc(100% - 44px);
 			}
 
 			iron-icon.completion-icon {
@@ -72,6 +73,7 @@
 				display: inline-flex;
 				flex: 1;
 				align-items: center;
+				max-width: 100%;
 			}
 
 			.course-name {
--- a/components/d2l-assessments-list-item.html
+++ b/components/d2l-assessments-list-item.html
@@ -79,6 +79,7 @@
 			.course-name {
 				display: inline;
 				word-wrap: break-word;
+				overflow-wrap: break-word;
 			}
 
 			.assessment-type {
--- a/components/d2l-assessments-list-item.html
+++ b/components/d2l-assessments-list-item.html
@@ -31,10 +31,11 @@
 			.assessment-info {
 				@apply --d2l-body-standard-text;
 				font-size: 12px;
-				display: flex;
-				flex-direction: row;
-				align-items: center;
+				/* display: flex; */
+				/* flex-direction: row; */
+				/* align-items: center; */
 				line-height: 1;
+				display: block;
 			}
 
 			.info-container {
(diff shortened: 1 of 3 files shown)


components/d2l-assessments-list.html
```diff
@@ -14,7 +14,7 @@
 			}
 
 			@media (max-width: 767px) {
-				d2l-assessments-list-item:nth-of-type(4) {
+				d2l-assessments-list-item:nth-of-type(n+4) {
```
tiffjiangd2l: I changed this because I believe that this is supposed to be hiding all items after and including the 4th item, and not just the fourth item? I can revert this change if the intention was just to hide the fourth item.

tiffjiangd2l: I changed this because I believe that this is supposed to be hiding all items after and including the 4th item, and not just the fourth item? I can revert this change if the intention was just to hide the fourth item.

CodeBaboon: There is a `slice` somewhere that limits the collection to max at 4.

tiffjiangd2l: Yeah that's what I meant. Previously it hid the 4th item but still showed the fifth, sixth etc. This change makes it so that it hides everything past the fourth.

CodeBaboon: I guess what I mean is that there should never be more than 4 in the collection. But this is probably better just in case, so it's all good.

components/d2l-assessments-list-item.html
```diff
@@ -74,8 +73,22 @@
...
+				max-width: 100%;
+			}
+
+			.course-name {
+				display: inline;
+				word-wrap: break-word;
+				overflow-wrap: break-word;
+			}
+
+			.assessment-type {
```
jstefaniuk-d2l: little tip for when you have multiple classes with the same data:
```
.date, .assessment-type {
     display: inline;
}
```

components/d2l-assessments-list-item.html
```diff
@@ -31,14 +31,13 @@
...
-				display: flex;
-				flex-direction: row;
-				align-items: center;
 				line-height: 1;
+				display: block;
 			}
 
 			.info-container {
 				display: block;
+				width: calc(100% - 44px);
```
jstefaniuk-d2l: Do we want to hardcode the 22px margins? Or would it be good to set a straight percent amount so that they shrink as you shrink the page? (probably a question for bob)

components/d2l-assessments-list.html
tiffjiangd2l: Ooooh I see, sorry for misunderstanding! Is that something I should fix here?

components/d2l-assessments-list-item.html
tiffjiangd2l: Thanks! :)

components/d2l-assessments-list-item.html
tiffjiangd2l: So the 44 px came from the hardcoded margins for the icon [here](https://github.com/Brightspace/d2l-upcoming-assessments-ui/blob/6245d755b53c13de2c7586941f8b74a66cdf9012/components/d2l-assessments-list-item.html#L58) (although I realize now that I forgot to account for rtl). But are you saying that it would be better to just set the width to e.g. 80%?

jstefaniuk-d2l: Actually... I misunderstood what this thing was... Carry on!

ChrisKraljevicD2L:
Has there been a release cut and pp's version bumped for this?

CodeBaboon:
not yet, I'll be on that in a bit

PR #71, opened 2017-10-27
Bump to new major of d2l-fetch-siren-entity-behavior

CodeBaboon:
We don't make use of any of the changed features in this repo but it keeps it to the same version as used by recent grades and the parent portal.


ChrisKraljevicD2L:
Is there a release with this PR's changes in it? I don't see any bower update in https://github.com/Brightspace/parent-portal-ui/pull/390/files

PR #73, opened 2017-10-31
Fail gracefully for failed activity requests

awikkerink:


CodeBaboon:
While I'm not totally opposed to this enhancement because you never know when something might fail and should it totally jack the page, I also feel like it is only fixing a symptom of a bigger issue. Why are we getting back links to invalid activity requests? It seems like the hm result should not send them, similar to how it would not send an action if the user didn't have permission to invoke it.

Surekh:
@CodeBaboon - @awikkerink - Did mention in slack that he will be working on API side changes. If this change is ok, can we get it through? One of our clients is planning to go live soon and this is blocking them.

CodeBaboon:
I'm undecided. I feel like we give a false sense of success in this case. "Here is all your work to do", except really a couple things that failed are missing. Can lead to loss of trust in the system.

Surekh:
Given that the current error makes the system unusable, and that we will be working on a API fix soon. I am willing to take the risk as a stop-gap solution. Is that ok? @CodeBaboon @awikkerink - Thoughts?. Also, the only known error is when user actually does not have access to a folder.

awikkerink:
Maybe if we check for 4xx errors?

CodeBaboon:
I think these known issues come back as 403s? Can't recall.

PR #74, opened 2017-10-31
backport: Fail gracefully for failed activity requests

CodeBaboon:
Fixes [DE28074 - 10.7.7 Upcoming Assignments are not loaded when one of the folders is not found](https://rally1.rallydev.com/#/157196228032d/detail/defect/169949292432)

NOTE: Branch name is incorrect, I crossed the streams. Sorry for confusion, but this is specifically for 10.7.7.

Couldn't cherry-pick the original because the `d2l-fetch-siren-entity-behavior` didn't exist pre-10.7.8. Will need a set of eyes/testing. Should be able to cherry-pick this one back to 10.7.6.


Surekh:
@CodeBaboon - I have a 1711 instance that I use for client demos. I can smoke test this, if you can get me a FRA endpoint.

CodeBaboon:
I've temporarily pointed https://github.com/Brightspace/parent-portal-ui/pull/397 to this branch. BrightspaceGithubReader should comment a fra endpoint if the build succeeds.

PR #78, opened 2017-11-06
Use custom 2 week window for activities usages

capajon:
Basically on the student selection screen, for the upcoming assessments count, we were pulling activities for the default date range provided by the user api call - so one week I believe.  This uses the custom date range used by the assessment widget - moved over to the already shared behavior.

This could definitely use some solid review and hopefully pre-merge testing.

### DIFF
--- a/behaviors/d2l-upcoming-assessments-behavior.html
+++ b/behaviors/d2l-upcoming-assessments-behavior.html
@@ -174,9 +174,13 @@
 
 		_getUserActivityUsages: function(userEntity, getToken, userUrl) {
 			var myActivitiesLink = (userEntity.getLinkByRel(this.HypermediaRels.Activities.myActivities) || {}).href;
+			var self = this;
 
 			if (myActivitiesLink) {
-				return this._fetchEntityWithToken(myActivitiesLink, getToken, userUrl);
+				return this._getCustomRangeAction(myActivitiesLink)
+					.then(function(actitiviesLink) {
+						return self._fetchEntityWithToken(actitiviesLink, getToken, userUrl);
+					});
 			}
 		},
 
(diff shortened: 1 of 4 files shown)


capajon:
Is it okay that this is impacted? http://search.dev.d2l/source/xref/GitHub/parent-portal-ui/src/components/overdue-assignments-alert.html#128

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -189,6 +193,57 @@
...
+
+			var self = this;
+
+			return this._fetchEntityWithToken(activitiesUrl, this.getToken, this.userUrl)
+				.then(function(activities) {
+					var currentTime = new Date();
+					var parameters = self._getCustomDateRangeParameters(currentTime);
+					var action = (activities.getActionByName(self.HypermediaActions.activities.selectCustomDateRange) || {});
+
+					self._selectCustomDateRangeAction = action;
```
capajon: This smells bad - a behavior probably shouldn't do this sort of thing, but I haven't had time to look at it closer/improve it.

ChrisKraljevicD2L: This property is used [here](https://github.com/Brightspace/d2l-upcoming-assessments-ui/blob/ecd44943e06a3b04af00ca13b20c40ca71c2deb1/components/d2l-upcoming-assessments.html#L336) to construct the action for a new date range without hypermedia-ing to get the action again. I'm kinda fuzzy on whether or not this is a bad smell too. I'm leaning towards this is OK but not ideal (since I have no clue how to get the action to that handler another way)

ChrisKraljevicD2L:
With respect to the overdue assignments alert we should check with Bob if two weeks of overdue stuff is cool since I believe he owns the design.

PR #83, opened 2017-11-28
Dates are not bold anymore

tiffjiangd2l:
I definitely left this out of my previous PR on purpose.

### DIFF
--- a/components/d2l-assessments-list-item.html
+++ b/components/d2l-assessments-list-item.html
@@ -59,10 +59,6 @@
 				height: 14px;
 			}
 
-			.date {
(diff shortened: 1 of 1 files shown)


Surekh:
@tiffjiangd2l - Thanks for noticing this. However, I wonder if this is an intentional design change. Can we confirm with Alex once before merging? I expect date to catch my eye when I see Upcoming work widget and that doesn't happen if the date isn't bold.

tiffjiangd2l:
Confirmed with Alex - the dates shouldn't be bolded :)

PR #84, opened 2017-12-05
US92034 upcoming work is now an immersive page

ChrisKraljevicD2L:
[Rally Story](https://rally1.rallydev.com/#/157196228032/detail/userstory/173954510288)

### DIFF
--- a/behaviors/d2l-upcoming-assessments-behavior.html
+++ b/behaviors/d2l-upcoming-assessments-behavior.html
@@ -1,7 +1,9 @@
 <link rel="import" href="../../polymer/polymer.html">
 <link rel="import" href="../../d2l-fetch-siren-entity-behavior/d2l-fetch-siren-entity-behavior.html">
 <link rel="import" href="../../d2l-hm-constants-behavior/d2l-hm-constants-behavior.html">
+<link rel="import" href="../../d2l-intl/d2l-intl.html">
 <link rel="import" href="./date-behavior.html">
+<link rel="import" href="./localize-behavior.html">
 
 <script src="https://s.brightspace.com/lib/siren-parser/6.0.0/siren-parser.js"></script>
 <script>
@@ -370,6 +372,7 @@
 		window.D2L.UpcomingAssessments.DateBehavior,
 		window.D2L.FetchSirenEntityBehavior,
 		window.D2L.Hypermedia.HMConstantsBehavior,
+		window.D2L.UpcomingAssessments.LocalizeBehavior,
 		upcomingAssessmentsBehaviorImpl
 	];
 
--- a/behaviors/d2l-upcoming-assessments-behavior.html
+++ b/behaviors/d2l-upcoming-assessments-behavior.html
@@ -259,14 +259,6 @@
 			return this._fetchEntityWithToken(userUrl, getToken);
 		},
 
-		__getFirstName: function(userEntity) {
-			var firstNameEntity = userEntity.getSubEntityByRel(this.HypermediaRels.firstName);
-			if (firstNameEntity) {
-				return firstNameEntity.properties.name;
-			}
-			return null;
-		},
-
 		_getInfo: function() {
 			this._showError = false;
 			var self = this;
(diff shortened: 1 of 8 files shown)


ChrisKraljevicD2L:
This guy and its [companion parent portal ui PR](https://github.com/Brightspace/parent-portal-ui/pull/448) are ready for review/test

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -1,7 +1,9 @@
 <link rel="import" href="../../polymer/polymer.html">
 <link rel="import" href="../../d2l-fetch-siren-entity-behavior/d2l-fetch-siren-entity-behavior.html">
 <link rel="import" href="../../d2l-hm-constants-behavior/d2l-hm-constants-behavior.html">
+<link rel="import" href="../../d2l-intl/d2l-intl.html">
```
CodeBaboon: `d2l-intl` is the new approved i18n thing, ja?

ChrisKraljevicD2L: shrug

This was included in `d2l-upcoming-assessments.html` and I need the stuff from it for `this.localize` in here.

components/d2l-all-assessments.html
```diff
@@ -0,0 +1,102 @@
...
+<link rel="import" href="../../d2l-link/d2l-link.html">
+<link rel="import" href="../../d2l-page-load-progress/d2l-page-load-progress.html">
+<link rel="import" href="../../d2l-typography/d2l-typography-shared-styles.html">
+<link rel="import" href="../behaviors/d2l-upcoming-assessments-behavior.html">
+<link rel="import" href="../behaviors/date-behavior.html">
+<link rel="import" href="../behaviors/localize-behavior.html">
+<link rel="import" href="d2l-all-assessments-list.html">
+<link rel="import" href="d2l-assessments-list.html">
+<link rel="import" href="d2l-date-dropdown.html">
+<script src="../../fetch/fetch.js"></script>
```
CodeBaboon: This feels nasty. I see we are doing it in recent-grades as well. I think it might be time to move to the bower version throughout the portal. GEMO for now though.

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -227,6 +255,99 @@
 			};
 		},
 
+		_getUser: function(userUrl, getToken) {
+			return this._fetchEntityWithToken(userUrl, getToken);
+		},
+
+		_storeUserInformation: function(userEntity) {
+			this._firstName = this.__getFirstName(userEntity);
```
CodeBaboon: Is this an unrelated fix that is thrown in with this PR? I like fixes but would in future prefer them seperate.

ChrisKraljevicD2L: Well that ain't used anywhere, away it goes! Whoops

components/d2l-all-assessments.html
ChrisKraljevicD2L: It looks like [parent portal](https://github.com/Brightspace/parent-portal-ui/blob/master/bower.json#L36) and [upcoming work](https://github.com/Brightspace/d2l-upcoming-assessments-ui/blob/master/bower.json#L17) use the same `fetch` version in their `bower.json`s. Is there a different `fetch` that yer referencing? Or is it how `fetch` is being included here?

CodeBaboon: Oh that's `fetch` fetch, not `d2l-fetch`, I got confused! That's just the polyfill. Do we need that at all? We usually don't include polyfills in components, we let the parent fra handle that if it feels it might be necessary.

ChrisKraljevicD2L: I'll strip it out and see what happens.

CodeBaboon: It's a polyfill so it's removal will only be noticeable on browsers that make use of it. So maybe our versions of Edge or an older-ish Safari.

PR #92, opened 2018-01-26
improve wrapping behaviour

margaree:
https://rally1.rallydev.com/#/157196228032/detail/defect/193096650252

Now looks like:
screen shot 2018-01-26 at 11 02 16 am

### DIFF
--- a/components/d2l-all-assessments-list-item.html
+++ b/components/d2l-all-assessments-list-item.html
@@ -94,7 +94,12 @@
 			.info-container {
 				padding-right: 10px;
 			}
-			
+
+			:host-context([dir="rtl"]) .info-container  {
+				padding-right: 0;
+				padding-left: 10px;
+			}
+
 			.meta-info {
 				display: inline-flex;
 				flex: 1;
--- a/components/d2l-all-assessments-list-item.html
+++ b/components/d2l-all-assessments-list-item.html
@@ -91,6 +91,10 @@
 				align-items: center;
 			}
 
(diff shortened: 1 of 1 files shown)


jstefaniuk-d2l:
Should there be any right padding in these? I always find it a bit bleh when text goes right to the border of bigger tiles like this

components/d2l-all-assessments-list-item.html
```diff
@@ -91,6 +91,10 @@
 				align-items: center;
 			}
 
+			.info-container {
+				padding-right: 10px;
```
CodeBaboon: Should have an RTL flip for this also.

ChrisKraljevicD2L:
I can test if there's no more review thingies

ChrisKraljevicD2L:
Testing Checklist
Sanity Checks
- [ ] Verify that unit tests and spec tests have been written, if appropriate
- [ ] Verify that unit tests, spec tests, and linting are all passing
- [ ] Verify that no code review comments are outstanding
- [ ] Verify that the code is behind a feature flag

Acceptance/Functionality
- [ ] Check that the acceptance criteria of the story has been fulfilled
- [ ] Check that related functionality has not been severely affected (smoke test)

Accessibility (UI)
- [ ] Verify that workflows can be completed via screen reader/keyboard
- [ ] Verify that workflows can be completed when the browser is zoomed to 200%

Internationalization (UI)
- [ ] Verify that lang terms are not hard coded

PR #97, opened 2018-02-13
Use siren-parser-import

ryantmer:
This allows for correct versioning of the script to avoid nasty surprises about which version is actually being fetched. Also removed direct references to the CDN script where they weren't needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,14 @@
 language: node_js
-node_js: 6
+node_js: node
 dist: trusty
 sudo: required
 before_script:
 - npm run test:lint:js
 - npm run test:lint:wc
-- export CHROME_BIN=/usr/bin/google-chrome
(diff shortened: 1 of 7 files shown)


ryantmer:
`bower link`ed this to PPUI yesterday and all appeared to work - we're not using any newish methods (like `getLinksByClasses`) in this repo, so really it doesn't overly care which version ^6.0.0 of the siren parser actually ends up on `window.D2L.Hypermedia.Siren.Parse`.

ryantmer:
In theory no, in practice yes - in particular, `d2l-recent-grades-ui` is pulling in something (I think `assignments-ui`?) that is still grabbing the 6.0.0 script, which is problematic for that repo, since it uses `getLinksByClasses`. So, I'll submit a PR to the few remaining repos and that should clear that up.

PR #99, opened 2018-02-27
US94716 Bump colors and hm constants

ChrisKraljevicD2L:
[Rally story](https://rally1.rallydev.com/#/157196228032/detail/userstory/199780077000)


CodeBaboon:
Did you say bump?
keef

ChrisKraljevicD2L:
I know that's keith richards but how does that connect to bump

PR #101, opened 2018-03-09
Meta text now splits onto two lines at mobile breakpoint

ChrisKraljevicD2L:


ChrisKraljevicD2L:
@CodeBaboon @ryantmer @margaree @jstefaniuk-d2l any reviews/objections for this guy?

components/d2l-all-assessments-list-item.html
```diff
@@ -143,6 +157,22 @@
 				.info-row {
 					display: none;
 				}
+
+				.row.meta-row {
+					flex-direction: column;
```
apalaniuk: Doesn't feel like much of a "row" anymore 😛 

Does setting `.meta-info` to 100% (as is already done) and `flex-wrap: wrap` not accomplish this without changing `flex-direction`?

components/d2l-all-assessments-list-item.html
```diff
@@ -193,26 +223,24 @@ <h3 class="assessment-title">[[assessmentItem.name]]</h3>
 						</div>
 					</div>
 				</div>
-				<div class="row">
-					<div class="spacer"></div>
-					<div class="activity-info">
-						<div class="info-container">
```
apalaniuk: Seems like there's a couple of CSS rules that could go if this is being removed

components/d2l-all-assessments-list-item.html
```diff
@@ -52,6 +52,15 @@
 				align-items: center;
 			}
 
+			.row.meta-row {
```
apalaniuk: nit: a `meta-row` is probably always a `row`

components/d2l-all-assessments-list-item.html
```diff
@@ -359,7 +387,7 @@ <h3 class="assessment-title">[[assessmentItem.name]]</h3>
 			if (canOpenActivityDetails) {
 				classNames = classNames + ' has-activity-details';
 			}
-			return classNames
+			return classNames;
```
apalaniuk: I wonder if removing `--fix` from linting makes sense, at least from any script run in CI, as it will actually mask existing issues

components/d2l-all-assessments-list-item.html
apalaniuk: Also, this is/was adding 10px of padding. From whence doth this spacing come in this PR? Or is there also an overall padding change?

components/d2l-all-assessments-list-item.html
ChrisKraljevicD2L: That's fair, I can rename the class to be something less misleading

The idea behind this rule is to have the two `div`s with this class on them stack at mobile widths but display together in one row at desktop widths. This seemed like the "simplest" way to accomplish that

components/d2l-all-assessments-list-item.html
ChrisKraljevicD2L: Whoops, I shall throw that back in. Did not mean to remove

components/d2l-all-assessments-list-item.html
apalaniuk: Yeah, nested flexboxes make things a bit more restrictive, so flipping between row/column makes sense

PR #105, opened 2018-04-16
DE29498: add touchend event listener

margaree:
https://rally1.rallydev.com/#/157196228032d/detail/defect/206949717816


jstefaniuk-d2l:
Code LGTM

PR #113, opened 2018-05-16
Hybridize

jstefaniuk-d2l:
Hybridize upcoming assessments, and change the `vaadin-date-picker` code to instead make use of the `d2l-date-picker`


.travis.yml
```diff
@@ -1,5 +1,5 @@
 language: node_js
-node_js: 9
+node_js: 8
```
CodeBaboon: Why 8?

bower.json
```diff
@@ -3,35 +3,52 @@
...
     "d2l-polymer-behaviors": "^1.0.0",
-    "d2l-offscreen": "^2.2.5",
-    "webcomponentsjs": "^0.7.0",
-    "d2l-colors": "^2.2.3",
+    "webcomponentsjs": "^1.1.0",
+    "d2l-colors": "^3.1.2",
     "d2l-icons": "^4.10.0",
-    "iron-iconset-svg": "^2.1.0"
+    "d2l-typography": "^6.0.0",
+    "iron-icon": "^2.0.0"
```
CodeBaboon: How come we still need all these resolutions? It's not super surprising in the 1.x variant but in the hybrid 2+ portion it would be nice to get rid of most/all of these.

components/d2l-assessments-list-item.html
```diff
@@ -227,11 +227,13 @@ <h3 class="assessment-title">{{assessmentItem.name}}</h3>
 		},
 
 		_computeCanOpenActivityDetails: function(assessmentItem, activityDetailsEnabled) {
-			return activityDetailsEnabled && assessmentItem.type === 'assignment' && !!assessmentItem.userActivityUsageHref;
+			return activityDetailsEnabled && assessmentItem && assessmentItem.type === 'assignment' && !!assessmentItem.userActivityUsageHref;
```
CodeBaboon: Might be time to add some line breaks in here for readability.

polymer.json
```diff
@@ -1,5 +1,6 @@
 {
   "lint": {
-    "rules": ["polymer-2-hybrid"]
+    "rules": ["polymer-2-hybrid"],
+    "filesToIgnore": ["index.html"]
```
CodeBaboon: how come?

jstefaniuk-d2l: This can actually be removed

bower.json
jstefaniuk-d2l: Yeah something upstream probably has a dependency on something old (d2l-demo-page perhaps?) We could maybe spend time trying to track down the culprit, but I'm not sure it's worth it atm (At least until we have this all going and can look for things without holding up the whole hybridization dealy)

.travis.yml
jstefaniuk-d2l: It was an arbitrary # between 6 and 9 and I was wondering if 9 was causing my issues from before

bower.json
CodeBaboon: Yeah I'm ok with that as long as we make a sweep through these afterwards. I think we really should be driving to a goal of almost 0 resolutions throughout our controlled dependency chain, with the future goal of then making sure it stays that way.

CodeBaboon:
Code seems reasonable.

PR #118, opened 2018-06-13
US96626 - Integrate with serge.io

apalaniuk:
Sort of gross in that build files are being committed. Will come back and do a second pass that does this during the build that publishes build resources to artifactory, just want to get this hooked into serge for now.


apalaniuk:
@CodeBaboon @ChrisKraljevicD2L old-ish PR with some localization process changes. A bit gross right now - will come back and automate the manual processes once we know what those should look like.

PR #125, opened 2018-07-26
Remove all the type-specific if chains

jstefaniuk-d2l:
I wanted to do this before starting on Discussions status badges, there should be no change in behavior


CodeBaboon:
Dev tested, LGTM

PR #130, opened 2018-08-09
Make sure to return endDate as string to be same as dueDate

CodeBaboon:
[DE30994 - Closed quizzes do not sort correctly in upcoming work list](https://rally1.rallydev.com/#/157196228032d/detail/defect/245133560576)

The `dueDate` is being returned as a string but the `endDate` was a date object. This led to incorrect comparisoning. I would have switched the `dueDate` to be a proper date to match `endDate` but that caused deeper problems with other places `new Date()`-ing the dueDate because these have traditionally been strings, so this was the safer fix.


ChrisKraljevicD2L:
I test

PR #131, opened 2018-08-23
fix problem with finding the activity link sometimes

jstefaniuk-d2l:
It was just taking the first link for discussions and quizzes, which was fine before the new pr reordered things

### DIFF
--- a/test/behaviors/d2l-upcoming-assessments-behavior.js
+++ b/test/behaviors/d2l-upcoming-assessments-behavior.js
@@ -94,7 +94,19 @@ describe('d2l upcoming assessments behavior', function() {
 		} else {
 			subEntities.push(getInstructions());
 		}
-		var activityClass = type === 'discussion' ? 'https://discussions.api.brightspace.com/rels/topic' : type;
+		var activityClass = type === 'discussion' ? 'topic' : type;
+		var rel;
+		switch (type) {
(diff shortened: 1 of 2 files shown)


ctwomey1:
Tests?

jstefaniuk-d2l:
Tests added

test/behaviors/d2l-upcoming-assessments-behavior.js
```diff
@@ -94,7 +94,19 @@ describe('d2l upcoming assessments behavior', function() {
 		} else {
 			subEntities.push(getInstructions());
 		}
-		var activityClass = type === 'discussion' ? 'https://discussions.api.brightspace.com/rels/topic' : type;
+		var activityClass = type === 'discussion' ? 'topic' : type;
+		var rel;
+		switch (type) {
```
ctwomey1: You might want to have a default case.  Siren subentities must have a rel.  You'd want your tests to have well formed data right?

ctwomey1: Might consider just having a guard statement or a default statement so that any activity not of discussion/assignment/quiz fails ish.

Side note might be a good idea at some point to use a siren entitiy builder instead of hand writing the siren.  https://github.com/Brightspace/node-siren-builder is one we maintain

PR #132, opened 2018-08-29
Updating Translations 1809 2018-8-29

BrightspaceGitHubReader:


CodeBaboon:
Failing because fork. Same dealio as [here](https://github.com/Brightspace/parent-portal-ui/pull/660#issuecomment-417080174). Code seems fine otherwise.

PR #133, opened 2018-09-06
Add content to upcoming assessments

jstefaniuk-d2l:
Nothing's deduped yet, but now things will show up (with the right icons!)
image

Right now there are some missing tier 3 icons that we need if we want to show tier 3 icons in the "all upcoming work" view, so we're just using tier2 there for now:
https://rally1.rallydev.com/#/157196228032/detail/userstory/251919894368?fdp=true


jstefaniuk-d2l:
Right now there are some missing tier 3 icons that we need if we want to show tier 3 icons in the "all upcoming work" view, so we're just using tier2 there for now

ChrisKraljevicD2L:
Looks good, I shall test as well

PR #135, opened 2018-09-13
DE31336 - Don't load module user activity usages into the widget

jstefaniuk-d2l:
We don't want to show modules, but we were loading them in anyways and that was causing things to crash

Needs:
https://git.dev.d2l/projects/CORE/repos/le/pull-requests/11940/overview
https://github.com/Brightspace/d2l-hm-constants-behavior/pull/65

### DIFF
--- a/behaviors/d2l-upcoming-assessments-behavior.html
+++ b/behaviors/d2l-upcoming-assessments-behavior.html
@@ -92,6 +92,9 @@
 			this._allTypes.forEach(function(typeString) {
 				var type = this._types[typeString];
 				var entities = usageList.getSubEntitiesByClass(type.userActivityUsageClass);
+				if (type.filterUsages) {
+					entities = entities.filter(type.filterUsages.bind(this));
+				}
 				usages = usages.concat(entities);
 			}.bind(this));
 			return usages;
--- a/behaviors/d2l-upcoming-assessments-behavior.html
+++ b/behaviors/d2l-upcoming-assessments-behavior.html
@@ -93,7 +93,7 @@
 				var type = this._types[typeString];
 				var entities = usageList.getSubEntitiesByClass(type.userActivityUsageClass);
 				if (type.usagePredicate) {
-					entities = entities.filter(type.usagePredicate.bind(this));
+					entities = entities.filter(type.usagePredicate);
 				}
 				usages = usages.concat(entities);
 			}.bind(this));
--- a/behaviors/d2l-upcoming-assessments-behavior.html
+++ b/behaviors/d2l-upcoming-assessments-behavior.html
@@ -92,8 +92,8 @@
 			this._allTypes.forEach(function(typeString) {
 				var type = this._types[typeString];
 				var entities = usageList.getSubEntitiesByClass(type.userActivityUsageClass);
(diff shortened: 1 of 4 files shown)


jstefaniuk-d2l:
Builds gonna fail until the HM PR gets in

behaviors/types-behavior.html
```diff
@@ -61,6 +61,9 @@
 							canOpen: false,
 							instructionsRel: this.HypermediaRels.Content.description,
 							userActivityUsageClass: this.HypermediaClasses.activities.userContentActivity,
+							filterUsages: function(userActivityUsage) {
```
omsmith: nit: this sounds like a verb to me, and I would expect it to work like:

```js
const usages = [1, 2, 3];
const filteredUsages = filterUsages(usages); // [2, 3]
```

Can I suggest `usagePredicate`?

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -92,6 +92,9 @@
 			this._allTypes.forEach(function(typeString) {
 				var type = this._types[typeString];
 				var entities = usageList.getSubEntitiesByClass(type.userActivityUsageClass);
+				if (type.usagePredicate) {
+					entities = entities.filter(type.usagePredicate.bind(this));
```
omsmith: ~nit: slightly concerned by the `this` usage (`this.HypermediaClasses.content.topic`) in the predicate and the binding here. How are we sure the `this` you're binding to here is the one the predicate is expecting?

jstefaniuk-d2l: Yeah, this would completely crash if the `this` wasn't the expected `this`

(the `.bind` on line 99 is the thing that binds to the `.this` for this polymer component)

jstefaniuk-d2l: Or you're saying it's possible that the predicate could expect a different `this` in the future?

jstefaniuk-d2l: I can't really see it using anything on the `this` object other than `HMConstants`, so I'm not sure if that's an issue

omsmith:
```js
							userActivityUsageClass: this.HypermediaClasses.activities.userContentActivity,
							usagePredicate: function(userActivityUsage) {
								return userActivityUsage && userActivityUsage.hasClass(this.HypermediaClasses.content.topic);
							},
```

My concern / confusion is that the `this`es between these two definitions are different. The first one is whatever it was during the definition of the property, and the one in the function is whatever `this` is within this separate behaviour? Maybe they're the same because of how things are wired up, but at least to _me_, that is not obvious.

I would personally be more comfortable with below, and then not `bind`ing it anywhere else, but you're the expert.
```js
							userActivityUsageClass: this.HypermediaClasses.activities.userContentActivity,
							usagePredicate: (function(userActivityUsage) {
								return userActivityUsage && userActivityUsage.hasClass(this.HypermediaClasses.content.topic);
							}).bind(this),
```

omsmith:
:heart:

PR #136, opened 2018-09-26
Updating Translations 1810 2018-9-26

BrightspaceGitHubReader:


lang/ar.json
```diff
@@ -1,26 +1,26 @@
...
-  "noUpcomingAssessments": "ليس لـ {userName} أي عمل مستحق في الأسبوعَين المقبلَين.",
-  "quiz": "الاختبار",
-  "today": "اليوم",
-  "tomorrow": "غدًا",
-  "viewAllAssignments": "عرض كل العمل"
+﻿{
+   "activityComplete" : "تم الإكمال",
+   "activityDueToday" : "تاريخ الاستحقاق اليوم",
+   "activityEnded" : "مغلق",
+   "activityEndsToday" : "Ends Today",
```
jeff-shepstone: This string remains in English in all of the translations. It doesn't look like this is an issue with the now-Engilsh string being added after DevDone. Is it possible that it's always had this value before connecting with Serge?

jstefaniuk-d2l: I think it may have been added in at the same time as serge (Merged like a day before the original serge PR got in, but wasn't part of the original serge changes)

scowing: This term was translated in the release/10.8.5 (1809 train) export/import. It looks like the PR that serge created was never merged (https://github.com/Brightspace/d2l-upcoming-assessments-ui/pull/132/files?utf8=%E2%9C%93&diff=unified&w=1)

scowing: Maybe that PR can be merged into this one?

PR #138, opened 2018-11-01
Lang build files

ChrisKraljevicD2L:


build/langterms/ar.html
```diff
@@ -14,18 +14,18 @@
...
 				"assignment": "الفرض",
 				"closeSimpleOverlayText": "العودة إلى لوحة المعلومات",
 				"closeSimpleOverlayTextMobile": "لوحة المعلومات",
-				"content": "Content",
+				"content": "المحتوى",
 				"currentPeriod": "{startDate} - {endDate}",
 				"discussion": "المناقشة",
-				"dueDateShort": "تاريخ الاستحقاق {dueDate}",
 				"dueDateLongImminent": "{prefix}: {dueDate}",
+				"dueDateShort": "تاريخ الاستحقاق {dueDate}",
```
CodeBaboon: I don't think we can just ditch the `{prefix}` here, can we?

PR #147, opened 2019-02-13
add instructionsClass option for instructions without rels

jstefaniuk-d2l:
Checklists will likely need this so I thought I'd throw up a quick PR


jstefaniuk-d2l:
not needed

PR #148, opened 2019-03-14
Add surveys/checklists

apalaniuk:
Some gross things in here. Once BfP is converted to Polymer 3, most of this logic should be centralized into a client SDK that eg. simplifies the activities domain. Possibly also some general logic simplification could be done. (Of course, all my awful code just has to be in a public repo..)

### DIFF
--- a/behaviors/d2l-upcoming-assessments-behavior.html
+++ b/behaviors/d2l-upcoming-assessments-behavior.html
@@ -88,18 +88,7 @@
 		},
 
 		_concatActivityUsageTypes: function(usageList) {
-			var usages = [];
-			this._allTypes.forEach(function(typeString) {
-				var type = this._types[typeString];
-				var entities = usageList.filter(function(usage) {
-					return usage.hasClass(type.userActivityUsageClass);
-				});
-				if (type.usagePredicate) {
-					entities = entities.filter(type.usagePredicate);
-				}
-				usages = usages.concat(entities);
-			}.bind(this));
-			return usages;
+			return usageList.filter(this._isSupportedType.bind(this));
 		},
 
 		_getActivityStatus: function(type, userActivityUsage, overdueUserUsages) {
@@ -360,7 +349,9 @@
 				activityEntities = activities.entities || [];
 			}
 
-			var activitiesContext = this._createNormalizedEntityMap(activityEntities);
+			var supportedActivities = activityEntities.filter(this._isSupportedType.bind(this));
+
+			var activitiesContext = this._createNormalizedEntityMap(supportedActivities);
 			var flattenedActivities = Array.from(activitiesContext.activitiesMap.values());
 
 			return self._hydrateActivityEntities(flattenedActivities)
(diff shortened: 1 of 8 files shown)


behaviors/types-behavior.html
```diff
@@ -27,6 +27,40 @@
 							noCompletion: false,
 							activityDetailsFeatureFlag: 'assignmentDetailsEnabled'
 						},
+						checklistItem: {
+							icon: function(assessmentItem) {
+								// checklist doesn't have a tier 3 icon
+								return assessmentItem.tier2IconKey || 'd2l-tier2:checklist';
```
jstefaniuk-d2l: `tier2IconKey` is only really a content concept at the moment, since in that case we need to figure out the icon from like 20 different sources

you can probably just do:
```
icon: function() {
   return 'd2l-tier2:checklist';
}
```

(also in the other places)

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -326,6 +354,145 @@
...
+					var parentActivitiesMap = activitiesContext.parentActivitiesMap;
+
+					// Normalize activity data prior to deduping; eg, some activities don't
+					// have a due date (surveys), while the content topic can
+					hydratedActivities.forEach(function(activity) {
+						if (parentActivitiesMap.has(activity.getLinkByRel('self').href)) {
+							var parentActivity = parentActivitiesMap.get(activity.getLinkByRel('self').href);
+
+							if (activity.getSubEntitiesByClass('due-date').length === 0) {
+								var parentDueDates = parentActivity.getSubEntitiesByClass('due-date');
```
jstefaniuk-d2l: `var parentDueDates = parentActivity.getSubEntityByClass('due-date');` may be good since we're just grabbing the first element anyways

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -326,6 +354,145 @@
...
+					var activitiesMap = activitiesContext.activitiesMap;
+					var parentActivitiesMap = activitiesContext.parentActivitiesMap;
+
+					// Normalize activity data prior to deduping; eg, some activities don't
+					// have a due date (surveys), while the content topic can
+					hydratedActivities.forEach(function(activity) {
+						if (parentActivitiesMap.has(activity.getLinkByRel('self').href)) {
+							var parentActivity = parentActivitiesMap.get(activity.getLinkByRel('self').href);
+
+							if (activity.getSubEntitiesByClass('due-date').length === 0) {
```
jstefaniuk-d2l: Would `hasEntityByClass('due-date')` be good here?

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -326,6 +354,145 @@
...
+
+			// Already-complete entities
+			var hydratedActivities = activityEntities
+				.filter(function(entity) {
+					return !entity.href;
+				});
+
+			var activityPromises = activityEntities
+				.filter(function(entity) {
+					return !!entity.href;
```
jstefaniuk-d2l: I bet this works entirely the same without coercing it to a bool

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -326,6 +354,145 @@
...
+				})
+				.map(function(entity) {
+					return self._fetchEntityWithToken(entity.href, self.getToken, self.userUrl)
+						.then(function(entity) {
+							return Promise.resolve(window.D2L.Hypermedia.Siren.Parse(entity));
```
jstefaniuk-d2l: I don't think you need to `Promise.resolve` this, simply calling `return` from the `then` should do the same thing iirc

jstefaniuk-d2l: Something's up here, since `_fetchEntityWithToken` is already supposed to be doing this: https://github.com/Brightspace/d2l-fetch-siren-entity-behavior/blob/a5cf7958165ca9f5f16bdda9e200b4a071ea66f5/d2l-fetch-siren-entity-behavior.html#L81

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -326,6 +354,145 @@
...
+			var parentActivitiesMap = new Map();
+			var allActivities = [];
+			var self = this;
+
+			activityEntities = activityEntities.map(function(entity) {
+				return window.D2L.Hypermedia.Siren.Parse(entity);
+			});
+
+			activityEntities.forEach(function(activity) {
+				var sirenActivity = window.D2L.Hypermedia.Siren.Parse(activity);
```
jstefaniuk-d2l: Looks like this is being done twice:
https://github.com/Brightspace/d2l-upcoming-assessments-ui/pull/148/files#diff-3481579d7de2f978e7455a222257b475R417

behaviors/d2l-upcoming-assessments-behavior.html
apalaniuk: Right, in practice the augmented functions weren't showing up/working. There's definitely a chance that getting the functionality there meant I went overboard, or that something (eg. `Object.assign` somewhere) is interfering and the real fix is there. Definitely something I need to revisit.

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -326,6 +354,145 @@
...
+				});
+		},
+
+		_createNormalizedEntityMap: function(activityEntities) {
+			var activitiesMap = new Map();
+			var parentActivitiesMap = new Map();
+			var allActivities = [];
+			var self = this;
+
+			activityEntities = activityEntities.map(function(entity) {
```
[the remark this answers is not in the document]
apalaniuk: Probably. I tend to stay away from passing the function as the argument, since we tend to get into cases where it's less obvious what's getting passed down call chains if they're expanded over time but, I'm fine with things that are shorter.

behaviors/types-behavior.html
apalaniuk: True, it can just be eg. `icon: 'checklist'` if this isn't a forward-compatibility thing and it will only ever be for content topics.

jstefaniuk-d2l: Watch out for that since there's no tier3 for this one: https://github.com/Brightspace/d2l-upcoming-assessments-ui/blob/897c92ede622e1a2828d917fc953af8b2f5dc824/components/d2l-all-assessments-list-item.js#L336

apalaniuk: Right, so, our code shouldn't make the assumption about tiers probably.

jstefaniuk-d2l: before, everything we needed had both a tier2 and tier3 icon so it wasn't a problem, now may be a good time to make it better for the tier3-challenged domains out there 😄

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -326,6 +330,173 @@
...
+						.then(window.D2L.Hypermedia.Siren.Parse);
+				});
+
+			return Promise.all(activityPromises)
+				.then(function(entities) {
+					entities.forEach(function(entity) {
+						hydratedActivities.push(entity);
+					});
+
+					return hydratedActivities;
```
jstefaniuk-d2l: may be able to just do `return hydratedActivities.concat(entities);`

behaviors/d2l-upcoming-assessments-behavior.html
```diff
@@ -326,6 +330,173 @@
...
+						if (parentActivitiesMap.has(activity.getLinkByRel('self').href)) {
+							var parentActivity = parentActivitiesMap.get(activity.getLinkByRel('self').href);
+
+							// There are cases where a content topic child activity (eg. a survey activity) doesn't
+							// have the same set of restrictions as the content topic itself. Because we only want to
+							// display one version of the same logical activity, we'll use the child activity,
+							// but ensure it has the superset of data from the content topic (due date).
+							// Since our data model is currently based on the LMS Siren entities,
+							// create and parse a new synthetic entity.
+							if (!activity.hasEntityByClass('due-date') && parentActivity.hasEntityByClass('due-date')) {
```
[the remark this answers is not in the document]
apalaniuk: In hindsight this should probably be deferred to a renderer that's intelligent about content activities - you'd still have to dedupe and everything, but you'd keep the content activity with child activity, and it would just figure out the correct date restrictions when creating the view data object...

PR #149, opened 2019-03-14
Add composed:true for events to escape the shadows

CodeBaboon:


apalaniuk:
I wasn't aware until now that this component has knowledge of concepts like a parent's immersive page, and specific pages within BfP. Seems like a good case for pulling this into the BfP repo.

CodeBaboon:
@apalaniuk It's not really ideal, nor is it really good. At some point the plan/theory is this should be a true widget that could appear in the LMS. If/when that time comes I guess we would need to deal with these sorts of things. I'd rather leave it outside BfP to encourage that, and less coupling.

jstefaniuk-d2l:
If we were to do that, we'd likely want a way to launch activity-details as a popup from this widget, perhaps a `parent-widget-suite` repo would make sense since our widgets are pretty coupled in functionality

PR #159, opened 2019-07-08
touchend makes it so you accidentally click the links while scrolling

jstefaniuk-d2l:
Don't quite know the history of why this is here, but removing it makes it so you don't trigger the link when trying to scroll


CodeBaboon:
This is the history: https://github.com/Brightspace/d2l-upcoming-assessments-ui/pull/105

jstefaniuk-d2l:
Tried it out locally on an android device and I don't think that fix is needed anymore

PR #160, opened 2019-07-30
use d2l-icon rather than iron icon

jstefaniuk-d2l:


components/d2l-all-assessments-list-item.js
```diff
@@ -124,7 +124,7 @@ $_documentContainer.innerHTML = `<dom-module id="d2l-all-assessments-list-item">
 				padding-right: 30px;
 			}
 
-			iron-icon.separator-icon {
+			d2l-icon.separator-icon {
 				opacity: 0.8;
 				width: 18px;
 				height: 18px;
```
dlockhart: So tier1 icons are already 18x18, so these 2 lines shouldn't be needed.

PR #162, opened 2020-07-10
Abort stale request chains

apalaniuk:
DE38539

This updates requests to use [AbortController](https://developer.mozilla.org/en-US/docs/Web/API/AbortController)/[AbortSignal](https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal) in order to terminate request chains on user context changes.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,6 @@
 language: node_js
 node_js:
 - 10.15.3
-dist: trusty
-sudo: required
 addons:
   chrome: stable
 before_script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ before_script:
 script:
 - set -e
 - polymer test --skip-plugin sauce
-- 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then polymer test --skip-plugin local; fi'
+# - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then polymer test --skip-plugin local; fi'
 - if [ "$TRAVIS_BRANCH" == "master" ]; then frauci-update-version; fi
 env:
   global:
--- a/behaviors/d2l-upcoming-assessments-behavior.js
+++ b/behaviors/d2l-upcoming-assessments-behavior.js
@@ -255,13 +255,15 @@ var upcomingAssessmentsBehaviorImpl = {
 
 		this.__getInfoRequest = this.__getInfoRequest
 			.then(() => {
-				this.__getInfoAbortController = new AbortController();
+				if (window.AbortController) {
+					this.__getInfoAbortController = new AbortController();
+				}
 
 				return this._fetchEntityWithToken({
 					link: userUrl,
 					getToken: getToken,
 					requestInit: {
-						signal: this.__getInfoAbortController.signal,
+						signal: (this.__getInfoAbortController || {}).signal,
 					},
 				});
 			})
@@ -278,7 +280,7 @@ var upcomingAssessmentsBehaviorImpl = {
 					userLink: userUrl,
 					getToken: getToken,
 					requestInit: {
-						signal: self.__getInfoAbortController.signal,
+						signal: (self.__getInfoAbortController || {}).signal,
 					}
 				});
 			})
@@ -293,14 +295,15 @@ var upcomingAssessmentsBehaviorImpl = {
 					getToken: getToken,
 				});
 			})
-			.catch(function(e) {
+			.then(function() {
+				self.__getInfoAbortController = null;
+			}, function(e) {
+				self.__getInfoAbortController = null;
+
 				if (!(e instanceof Error) || e.name !== 'AbortError') {
 					self._showError = true;
 					self._firstName = null;
 				}
-			})
-			.finally(() => {
-				self.__getInfoAbortController = null;
 			});
 
 		return this.__getInfoRequest;
--- a/behaviors/d2l-upcoming-assessments-behavior.js
+++ b/behaviors/d2l-upcoming-assessments-behavior.js
@@ -37,9 +37,16 @@ var upcomingAssessmentsBehaviorImpl = {
 		_periodEnd: String
 	},
 
-	_getOrganizationRequest: function(userActivityUsage, getToken, userUrl) {
+	_getOrganizationRequest: function(userActivityUsage, getToken, userUrl, abortController) {
 		var organizationLink = (userActivityUsage.getLinkByRel(Rels.organization) || {}).href;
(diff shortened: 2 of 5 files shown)


behaviors/d2l-upcoming-assessments-behavior.js
```diff
@@ -228,8 +242,31 @@ var upcomingAssessmentsBehaviorImpl = {
 	_getInfo: function() {
 		this._showError = false;
 		var self = this;
+		var userUrl = this.userUrl;
```
apalaniuk, line 245: Ideally we'd make these methods more functional so that they always use the initial context throughout the request chain, though aborting is still desirable for bandwidth/HTTP1 connection reasons.

behaviors/d2l-upcoming-assessments-behavior.js
```diff
@@ -37,9 +37,16 @@ var upcomingAssessmentsBehaviorImpl = {
...
 	},
 
-	_getOrganizationRequest: function(userActivityUsage, getToken, userUrl) {
+	_getOrganizationRequest: function(userActivityUsage, getToken, userUrl, abortController) {
 		var organizationLink = (userActivityUsage.getLinkByRel(Rels.organization) || {}).href;
-		return this._fetchEntityWithToken(organizationLink, getToken, userUrl);
+		return this._fetchEntityWithToken({
+			link: organizationLink,
+			userLink: userUrl,
+			getToken: getToken,
```
jameswklassen, line 45: Can this be written as just `getToken` since the key and property are named the same?

behaviors/d2l-upcoming-assessments-behavior.js
```diff
@@ -434,7 +506,7 @@ var upcomingAssessmentsBehaviorImpl = {
 	/*
 	* On success, all activities, with linked subentities hydrated
 	*/
-	_hydrateActivityEntities: function(activityEntities) {
+	_hydrateActivityEntities: function(activityEntities, getToken, userUrl, abortController) {
```
jameswklassen, line 509: Even though it seems like we only use `abortController.signal`, I'm assuming it's best practice to always pass around the entire controller?

behaviors/d2l-upcoming-assessments-behavior.js
apalaniuk: [This ESLint rule prevents that](https://github.com/Brightspace/d2l-upcoming-assessments-ui/blob/master/.eslintrc.json#L6), but it was added [for a reason which is almost certainly no longer true](https://github.com/Brightspace/d2l-upcoming-assessments-ui/pull/62#discussion_r143088443), but I'm not sure about other contexts it's used in. Would be nice to dump that (and a lot of other _terrible_ things) when it's flipped over to LitElement, though.

behaviors/d2l-upcoming-assessments-behavior.js
apalaniuk: I have no problem flipping that over to just accepting an AbortSignal. :+1:

behaviors/d2l-upcoming-assessments-behavior.js
jameswklassen: Oh wow, 2017 😅 . Makes sense knowing the context – I've just been yelled at by many linters for not doing this.

behaviors/d2l-upcoming-assessments-behavior.js
jameswklassen: No strong feelings, especially since I don't know the full context/use of this component. Just something I noticed 🙂

apalaniuk: Too late!

behaviors/d2l-upcoming-assessments-behavior.js
```diff
@@ -290,7 +296,7 @@ var upcomingAssessmentsBehaviorImpl = {
 				return self._loadActivitiesForPeriod({
 					activitiesEntity: activitiesEntity,
 					dateObj: new Date(),
-					abortController: self.__getInfoAbortController,
+					abortSignal: (self.__getInfoAbortController || {}).signal,
```
jameswklassen, line 299: Is there a reason to use `(obj || {}).prop` syntax vs. `(obj?.prop)`? Is this in a similar compatibility vein as the linting rule?

jstefaniuk-d2l: `?.` isn't fully out yet, it's an option you currently have to turn on in the transpiler. Folio has it on, but parent portal currently doesn't

jameswklassen: 😞

apalaniuk: Yeah, was 50/50 on calling this out - I certainly prefer features like optional chaining and I push for them in our app code, and for our specific uses, we do transpile our dependencies/could flip it on easily where it's disabled, but it's safer for the moment to use the existing style since this is/may be a dependency of other things right now, and I'm not sure of the build process in all of those places. (JS-only plugins similar to [@typescript-eslint/prefer-optional-chain](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/prefer-optional-chain.md) are just starting to appear now, though not supported natively by ESLint, and my preference is to enable those once it's deemed safe)

.travis.yml
```diff
@@ -11,7 +9,7 @@ before_script:
 script:
 - set -e
 - polymer test --skip-plugin sauce
-- 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then polymer test --skip-plugin local; fi'
+# - 'if [ "$TRAVIS_PULL_REQUEST" != "false" ]; then polymer test --skip-plugin local; fi'
```
apalaniuk, line 12: Temporarily disabled - it's been quite a while since CI ran on this, so I think there's reconfig of Sauce required for the repo.

components/d2l-all-assessments.js
```diff
@@ -106,12 +109,45 @@ Polymer({
...
-			return this._loadActivitiesForPeriod(this.__activitiesEntity, e.detail.date)
+			if (this.__dateChangeAbortController) {
+				this.__dateChangeAbortController.abort();
+			}
+
+			if (!this.__dateChangeActivityLoadRequest) {
+				this.__dateChangeActivityLoadRequest = Promise.resolve();
+			}
+
+			const ctx = {
```
jstefaniuk-d2l, line 120: Any reason to have this object right now?

apalaniuk: This was initially part of propagating the initial context, but that was reverted a while ago, so I can just use these directly. :+1:

PR #172, opened 2021-04-30
Es6 behavior

jstefaniuk-d2l:


behaviors/d2l-upcoming-assessments-behavior.js
```diff
@@ -125,81 +124,66 @@ var upcomingAssessmentsBehaviorImpl = {
...
-		var requests = [];
+		const overdueUserUsages = this._concatActivityUsageTypes(overdueUserActivityUsages);
+		const supportedUserUsages = this._concatActivityUsageTypes(userActivityUsages);
 
-		var overdueUserUsages = this._concatActivityUsageTypes(overdueUserActivityUsages);
-		var supportedUserUsages = this._concatActivityUsageTypes(userActivityUsages);
+		const requests = supportedUserUsages.map(async(userActivityUsage) => {
+			const organizationRequest = this._getOrganizationRequest.call(this, userActivityUsage, getToken, userUrl, abortSignal);
+			const activityRequest = this._getActivityRequest.call(this, userActivityUsage, getToken, userUrl);
+			const userActivityUsageHref = userActivityUsage.getLinkByRel('self').href;
```
EvanLemieux, line 133: This is a lot cleaner. I like this.

behaviors/d2l-upcoming-assessments-behavior.js
```diff
@@ -217,23 +201,22 @@ var upcomingAssessmentsBehaviorImpl = {
 	},
 
 	_getCustomRangeAction: function(activitiesEntity, dateObj) {
-		var self = this;
```
EvanLemieux, line 220: Was this ever a recommended pattern? This seems so foreign to me lol

jstefaniuk-d2l: Yeah, back before `() => {}` was a thing, the `this` in anonymous functions was different from the context you created them in, so you would either have to `.bind(this)` them or define a `self` var for referencing instead

EvanLemieux: Ohhh yeah that makes a lot of sense. I know JS devs always joked about not knowing what `this` is, but it really didn't seem that bad. I only got into it after ES6 though so I guess the before times were even worse lol

EvanLemieux (approved):
This all seems pretty legit. LGTM :+1:

jstefaniuk-d2l:
I may hold off on merging this since it may be more of a "transition" step to something newer

PR #176, opened 2021-06-09
Remove overlay styling and langterms as component does not use simple-overlay

annabelsegalld2l:
It appears the component did use the d2l-simple-overlay component back in 2017 or so, but it has since been deleted and the unused langterms and styling have been left.

Here is the PR that deleted the overlay component: https://git.dev.d2l/projects/GIT/repos/d2l-upcoming-assessments-ui/commits/81c295abda668c7f19ab94158ac70def95352cce#components/d2l-upcoming-assessments.html


components/d2l-upcoming-assessments.js
```diff
@@ -204,12 +162,6 @@ Polymer({
 		'_onApiConfigChanged(userUrl, getToken)'
 	],
 
-	ready: function() {
-		var mql = window.matchMedia('(max-width: 767px)');
```
annabelsegalld2l, line 208: All of the listening for mobile breakpoints was to set the closeSimpleOverlayAltText, which is no longer used.

build/langterms/ar.js
```diff
@@ -15,8 +15,6 @@ window.D2L.UpcomingAssessments.LangArBehavior = {
 		"activityExempted": "معفى",
 		"activityOverdue": "تم تجاوز تاريخ الاستحقاق",
 		"assignment": "الفرض",
-		"closeSimpleOverlayText": "العودة إلى لوحة المعلومات",
```
annabelsegalld2l, line 18: Deleting these langterms as they are no longer used.

jstefaniuk-d2l (approved):

PR #154, opened 2019-04-29
[WIP] Surveys/Checklists changes in p3 now

ChrisKraljevicD2L:

PR #178, opened 2021-06-22
Langterm updates

ryantmer:
Follow-up to #167 - suppose I could have just added these to that PR, but... I didn't think of that, ok


lang/en.json
```diff
@@ -6,6 +6,7 @@
   "activityExempted": "Exempted",
   "activityOverdue": "Overdue",
   "assignment": "Assignment",
+  "checklistItem": "Checklist Item",
```
ryantmer, line 9: Mentioned in the commit, but these terms were supposed to be added in #154, but they were accidentally added to `build/langterms/en.js` rather than `lang/en.json`. They've caused a few problems since then, but pretty sure this will get things working again (and will make surveys/checklist items display correctly in BfP in other languages, too).

(There's no associated change to `build/langterms/en.js` here as they are already there.)

build/langterms/fr-ON.js
```diff
@@ -14,7 +14,7 @@ window.D2L.UpcomingAssessments.LangFrONBehavior = {
 		"activityEndsToday": "Fin aujourd'hui",
 		"activityExempted": "Exempté",
 		"activityOverdue": "En retard",
-		"assignment": "Tâche",
```
jstefaniuk-d2l, line 17: In this case we want `Tâche` https://github.com/Brightspace/d2l-upcoming-assessments-ui/pull/175

ryantmer: Ah yep, didn't even notice this was fr-ON. FWIW, the term was incorrect in `lang/fr-ON.json`, so I've fixed that, which also fixed this since this is just the build files.

jstefaniuk-d2l (approved):

build/langterms/fr-ON.js
jstefaniuk-d2l: Yeah I think getting rid of the langterm build step would reduce a lot of these errors for us, maybe a dragonslayer/slacktime task

ryantmer: Sort of in-progress, thanks to Lit-ification!

PR #179, opened 2021-06-29
US128094 Use HMC for View All Work screen

ryantmer:
Requires https://github.com/BrightspaceHypermediaComponents/foundation-components/pull/261

This is a conversion of the `d2l-all-assessments` component to instead use the non-collapsed version of the HMC `d2l-w2d-collections` component, plus a date selector that changes the date of which activities are currently showing the W2D component.

This required some ancillary changes here and there, but was a fairly straightforward replacement. Once we are no longer using the`d2l-upcoming-assessments` component (which has already been replaced by the collapsed W2D component in BfP), we can likely clean out a bunch of stuff from this repo, as it will effectively just be this `d2l-all-assessments` component - could even consider moving it back into the `parent-portal-ui` repo.

https://rally1.rallydev.com/#/378960033600d/custom/382879628416?detail=%2Fuserstory%2F604192818200&fdp=true


components/d2l-all-assessments.js
```diff
@@ -1,167 +1,94 @@
...
+		super();
+		this._setDates(new Date());
+	}
 
-	_formatPeriodText: function() {
-		var startDateDate = new Date(this._periodStart);
-		var endDateDate = new Date(this._periodEnd);
+	render() {
+		return html`
+			<d2l-input-date
```
ryantmer, line 52: I've realized I forgot to put a `label` on this component, and then further have realized that we maybe want some sort of design input for this date selector. What we were doing previously won't work here, so I'll follow up with Fisayo.

components/d2l-all-assessments.js
```diff
@@ -1,167 +1,94 @@
...
-			endDateString = this.formatDate(endDateDate, { format: 'monthDay' });
+	_onDateChanged(e) {
+		if (e.target.value) {
+			this._setDates(new Date(e.target.value));
 		}
+	}
 
-		this._currentPeriodText = this.localize('currentPeriod', 'startDate', startDateString, 'endDate', endDateString);
-	},
+	_setDates(selectedDate) {
```
ryantmer, line 81: This logic is mostly stolen directly from `d2l-upcoming-assessments-behavior._getCustomDateRangeParameters`, but that file will be dead soon enough so not much point in sharing this. (It's not 100% identical anyway due to `currentTime` being a thing here.)

ryantmer (commented):
Note that the change to the component here is a replacement, so the diff is pretty useless. Also taking a look at the lint/tests, not sure what this cryptic ESLint error means, nor how I would have introduced it here 🤔

components/d2l-all-assessments.js
ryantmer: Fixed now, just added a "Choose Date:" label per Fisayo's recommendation.

build/langterms/cy-gb.js
```diff
@@ -15,8 +15,8 @@ window.D2L.UpcomingAssessments.LangCygbBehavior = {
 		"activityExempted": "Wedi’i eithrio",
 		"activityOverdue": "Yn hwyr",
 		"assignment": "Aseiniad",
-		"closeSimpleOverlayText": "Yn ôl i’r Dangosfwrdd",
```
ryantmer, line 18: These are old unused terms, just hadn't been removed I think.

build/langterms/ar.js
```diff
@@ -15,6 +15,8 @@ window.D2L.UpcomingAssessments.LangArBehavior = {
 		"activityExempted": "معفى",
 		"activityOverdue": "تم تجاوز تاريخ الاستحقاق",
 		"assignment": "الفرض",
+		"checklistItem": "Checklist Item",
```
ryantmer, line 18: `checklistItem` and `survey` were missing from all languages except English, so I suspect they would have just fallen back to English? Either way, figured I'd fix them while I'm here.

build/langterms/en.js
```diff
@@ -10,17 +10,18 @@ window.D2L.UpcomingAssessments.LangEnBehavior = {
 	'en': {
 		"activityComplete": "Complete",
 		"activityDueToday": "Due Today",
-		"activityEndsToday": "Ends Today",
```
ryantmer, line 13: This file was _almost_ alphabetically-ordered - now it is 😄

polymer.json
```diff
@@ -1,6 +1,7 @@
 {
   "npm": true,
   "lint": {
-    "rules": ["polymer-3"]
+    "rules": ["polymer-3"],
+    "filesToIgnore": ["components/d2l-all-assessments.js"]
```
ryantmer, line 5: Not a Polymer element anymore - realistically all the Polymer stuff will disappear soon anyway, but this works as a stepping stone to that.

package.json
```diff
@@ -16,12 +16,13 @@
   },
   "scripts": {
     "build": "gulp build",
-    "test": "npm run test:lint:js && npm run test:lint:wc && npm run test:wct",
-    "test:local": "npm run test:lint:js && npm run test:lint:wc && npm run test:wct:local",
-    "test:lint:js": "eslint . --ext .js,.html",
-    "test:lint:wc": "polymer lint",
-    "test:wct": "polymer test -p",
-    "test:wct:local": "cross-env LAUNCHPAD_BROWSERS=chrome polymer test -p --skip-plugin sauce"
```
ryantmer, line 24: fun fact, pretty sure nothing is actually running the tests in Sauce here, despite all the config still being here. This PR is big enough though, so not going to tackle that here. (Not even sure if this repo will continue to exist once we're fully over to using W2D stuff - at that point, might make sense to move `d2l-all-assessments` back into the PPUI repo, since AFAIK nothing else here will be useful anymore.)

test/index.html
```diff
@@ -17,8 +17,7 @@
 				'./components/d2l-assessments-list.html',
 				'./components/d2l-assessments-list-item.html',
 				'./components/d2l-all-assessments-list-item.html',
-				'./components/d2l-date-dropdown.html',
-				'./components/d2l-all-assessments.html'
```
ryantmer, line 21: These tests are now run by `web-test-runner` rather than `web-component-tester`, which is old and crufty.

ryantmer (commented):
Hooray, fixed now! Actually ready for review - sorry for the false start.

jstefaniuk-d2l (commented):
Needs a major version bump I think, otherwise code LGTM

After that we can merge this and then test the BFP PR that bumps the version if that works for you?

jstefaniuk-d2l (approved):

PR #181, opened 2021-07-09
US128094 Fix date handling

ryantmer:
Followup/fix for #179.

There were two issues here causing weird behaviour with the date selector. First, by setting the `value` on the selector we create some strange cyclical behaviour, exacerbated by the problem outlined below. To avoid this, but to also avoid a blank date selector initially, we just set the date to "today" via a throwaway value. Once a date has been selected by a user, at that point it will just display the selected date.

The second issue was mostly apparently when selecting a date at the beginning of end of a week (i.e. Sunday or Saturday), as the offsets along with the cyclical behaviour mentioned above meant the calendar would go back a week every time you opened it. This was solved by adding the offset to the date value the `d2l-date-input` gives us, then working with that. This has the effect of turning the midnight-UTC the selector gives us into an equivalent datetime that is midnight-UTC but in the local timezone.

### DIFF
--- a/components/d2l-all-assessments.js
+++ b/components/d2l-all-assessments.js
@@ -17,7 +17,9 @@ class D2LAllAssessments extends LocalizeDynamicMixin(HypermediaStateMixin(LitEle
 			currentTime: { type: String, attribute: 'current-time' },
 			startDate: { type: String, attribute: 'start-date' },
 			endDate: { type: String, attribute: 'end-date' },
-			_myActivitiesHref: { type: String, observable: observableTypes.link, rel: rel.myActivities, prime: true }
+			_initialDateToDisplay: { type: String },
+			_myActivitiesHref: { type: String, observable: observableTypes.link, rel: rel.myActivities, prime: true },
+			_timezoneOffsetMinutes: { type: Number }
 		};
 	}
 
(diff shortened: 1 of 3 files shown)


components/d2l-all-assessments.js
```diff
@@ -81,14 +86,34 @@ class D2LAllAssessments extends LocalizeDynamicMixin(HypermediaStateMixin(LitEle
...
-		const dayOfWeek = selectedDate.getDay();
-		const beginningOfWeek = new Date(selectedDate.setDate(selectedDate.getDate() - dayOfWeek));
-		beginningOfWeek.setHours(0, 0, 0, 0);
+		// The date selector provides us with a datetime of midnight UTC of the selected day - but
+		// really, the user thinks they're selecting a datetime of midnight of the selected day in
+		// their local timezone. Because of this, we need to add their offset to this datetime so
+		// that we're still dealing in UTC datetimes, but that have been shifted by their local
+		// offset.
+		const localSelectedDate = new Date(selectedDate);
+		localSelectedDate.setMinutes(localSelectedDate.getMinutes() + this._timezoneOffsetMinutes);
```
ryantmer, line 95: I'm going to be completely honest - I don't really understand why this _wasn't_ an issue before 🤔 However, I've validated that this works for any date I give it. I'd love to write tests for this kind of behaviour, but that would involve mocking `Date` itself, which... I'm not sure if that's possible? I could write tests that would pass in CI (based on whatever timezone Travis runs in), but then they'd be pretty much guaranteed to fail locally :(

ryantmer:
I'm realizing it's kind of hard to test these components by themselves - does it make more sense to just merge this and update https://github.com/Brightspace/parent-portal-ui/pull/928 to include the new release this creates?

jstefaniuk-d2l (approved):

PR #182, opened 2021-07-19
DE44391 Fix date range clobbering behaviour

ryantmer:
See [this Slack thread](https://d2l.slack.com/archives/C01BHPXN2TE/p1626296500060800) for a bunch more discussion and investigation, but the short version here is that when we create the `d2l-w2d-collections` element in the View All Work view, it ends up being an element with the same `href` as the `d2l-w2d-collections` element created by the main W2D widget in BfP. The result of this is that both elements share the same HMC state, meaning changes to one (such as changing the date range) results in changes to the others. This means changing the date of the W2D widget in View All Work would _also_ change the date range of the main W2D widget, which we don't want.

This is a pretty simple resolution to this, which is simply to add a random query parameter - this does result in cache-busting behaviour, but not a huge deal, as we're ultimately going to be re-fetching things anyway due to the different date range.


adamsk34 (approved):
Interesting solution

ryantmer:
Hahah yeah, credit to Zina honestly - it's a bit hackish, but ultimately a non-hackish solution would require some significant changes to the W2D components and/or deep-down HMC state stuff, sooo.... hack.

components/d2l-all-assessments.js
```diff
@@ -18,7 +18,21 @@ class D2LAllAssessments extends LocalizeDynamicMixin(HypermediaStateMixin(LitEle
...
+				observable: observableTypes.link,
+				rel: rel.myActivities,
+				prime: true,
+				method: href => {
+					// This appends a random query parameter to the myActivitiesHref, to prevent
+					// the d2l-w2d-collections element we're creating here from sharing HMC state
+					// with any other ones on the page (resulting in them clobbering each others'
+					// date ranges)
+					const url = new URL(href);
+					url.searchParams.append('rand', Math.random());
```
jstefaniuk-d2l, line 32: Any chance we can make this a constant (`namespace: 'all-work'` or something)? so maybe it caches with itself?

ryantmer: This only gets called/run once anyway - this is only used when we do the first fetch to get the empty collection after we first create the element. If you switch weeks, we don't re-call this (and in fact even if you close the overlay and reopen it, we don't refetch it). Unless you mean caching across page loads? Is that a thing we do?

jstefaniuk-d2l: what if you switch students, and then back?

jstefaniuk-d2l: Also sending a random value would perhaps cause the server to not cache it?

jstefaniuk-d2l: Although the most significant caching for this particular route is done at the DB level, but there may be some caching higher up

jstefaniuk-d2l (approved):

components/d2l-all-assessments.js
ryantmer: Ahh yes, didn't think of switching back - that does indeed result in a new `rand` value. I don't think the response is cached at the URL level anyway, buuuut you are correct in that there's no need to make this completely random - really, all we need to do is differentiate it from the other `d2l-w2d-collections` element (created by the main W2D widget). Updated to just use a constant value, you've convinced me! 😄

PR #185, opened 2021-08-04
[BfP] Add loading shimmer to All Work

JayArora1997:
This PR is for the previous view all work that will stay around for next 2-3 releases. Therefore, the difference in the files is huge as this has now been converted to lit element and use new w2d component. We will again have to backport the lit changes when we release the new view all work page.

Testing instructions:
1. update the app config in a sitelord instance.
2. Go to student dashboard in bfp.
3. Select "view all work" in the w2d widget.

Before:

image
image

After:
image
image

https://rally1.rallydev.com/#/378960033600d/custom/382879628416?detail=%2Fuserstory%2F516663540244&fdp=true?fdp=true


components/d2l-all-assessments.js
```diff
@@ -1,136 +1,185 @@
...
-	static get localizeConfig() {
-		return {
-			importFunc: async lang => (await import(`../build/langterms/${lang}.js`)).default
-		};
-	}
+		<div class="period-selection-container">
+			<d2l-date-dropdown current-period-text="[[_currentPeriodText]]" locale="[[locale]]">
+			</d2l-date-dropdown>
+		</div>
+		<template is="dom-if" if="[[_isLoading]]">
```
JayArora1997, line 59: The only change here is the loading spinner on the top of old polymer file.

jameswklassen (approved):
code LGTM

behaviors/d2l-upcoming-assessments-behavior.js
```diff
@@ -22,6 +22,10 @@ var upcomingAssessmentsBehaviorImpl = {
 			type: Boolean,
 			value: false
 		},
+		_isLoading: {
```
jstefaniuk-d2l, line 25: do we ever have to re-set this to true?

JayArora1997: On re-render it grabs the initial true value which is further set to false once things are loaded. I don't think so but curious if we have to.

jstefaniuk-d2l: If we set it to false here: https://github.com/Brightspace/d2l-upcoming-assessments-ui/pull/185/files/e6528fa10a7e0fef3ff7e868bda69422ddcda242#diff-b47381e78b3f705b27559125de5b484177d50d2d44ab5ce1e659e63288253195R308 I think that it would allow us to get the spinner whenever it reloads (which it may do when switching students?)

jstefaniuk-d2l (approved):
code LGTM

PR #194, opened 2022-02-02
[Backport] Backport translations and bugfixes to v5

jameswklassen:


marcsoiferman (approved):
Looks like a backport

wct.conf.json
```diff
@@ -12,23 +12,18 @@
...
         },
         {
           "browserName": "safari",
-          "platform": "OS X 10.13",
-          "version": ""
-        },
-        {
-          "browserName": "microsoftedge",
-          "platform": "Windows 10",
-          "version": ""
```
jadasc, line 31: Are these changes in `wct.conf.json` necessary for v5?

jameswklassen: I'm also backporting the GitHub actions stuff so that the two CI pipelines are identical, so I think this change makes sense.

jameswklassen: Also, the old CI was never using these browsers, since it was running 'local' tests (i.e. not the defined saucelabs browser)

jadasc: Were the changes to this file originally surfaced by an automatic dependency update to `package-lock.json`? Or was that a different repo/problem?

jadasc (approved):

wct.conf.json
jameswklassen: discussed offline

